PR #1, opened 2017-04-05
Create Server Restructuring.md

tristantarrant:


Server Restructuring.md
```diff
@@ -0,0 +1,65 @@
+Infinispan Server Restructuring
+===============================
+
+Status quo
+----------
+
+Infinispan Server is currently based on the full WildFly distribution.
+It is composed of the following additional components:
+
+  * a JGroups subsystem which is mostly aligned with the WildFly with some additions (e.g. SASL integration)
```
slaskawi: and `KUBE_PING`

Server Restructuring.md
```diff
@@ -0,0 +1,65 @@
...
+Redesign proposals
+------------------
+
+For future development we want to pursue two paths: one which maintains compatibility with the current situation but resolves some of the above disadvantages and a more aggressive one tailored for containers
+
+Plan A: Base the server on WildFly Core
+
+   * Make the dependency on the transactions and datasource subsystems optional by providing our own internal implementations by directly integrating with Narayana and HikariCP
+   * Still work with the transactions/datasource subsystem if they are present (to support the WildFly layering)
+   * Need to still support deployable JDBC drivers
```
slaskawi: What's the motivation behind this? Maybe we could put the most popular drivers into the modules and ask users to put others inside a deployable cache stores?

Server Restructuring.md
```diff
@@ -0,0 +1,65 @@
...
+   
+Plan B: Create an uber-jar slim server
+
+   * Add some bootstrapping code
```
slaskawi: After thinking about this for a while, maybe we could bootstrap EmbeddedCacheManager with additional features (such as REST and HotRod server)?

Server Restructuring.md
```diff
@@ -0,0 +1,65 @@
...
+   * Add some bootstrapping code 
+     * no need for WildFly Swarm or Spring Boot
+   * Configuration
+     * use the embedded Infinispan schema
+     * configure endpoints and other components via the global modules API
+     * persist configuration changes using the existing configuration serializer
+     * add runtime clustered configuration capabilities (store configs in an internal cache)
+   * Security
+     * Depend on Elytron for LDAP, Properties, Certificates, Kerberos, KeyCloak, etc integration
+   * Custom user deployments
```
slaskawi: I'm not sure about this point here. We shouldn't support deployment but rather scan the classpath and bootstrap custom marshallers/filters/cache stores when they are discovered. As you mentioned, we will be using Uber Jar, so we won't have many elements to scan.

tristantarrant: These are not like the dynamic deployments of a traditional application server. It is mostly an 'ext' directory which is added to the classpath of the server. Nothing special here, but we need to ensure that we are consistent with the discovery method (META-INF/services/* usually)

Server Restructuring.md
tristantarrant: This is what we had before: https://github.com/infinispan/infinispan/blob/5.2.x/server/core/src/main/scala/org/infinispan/server/core/Main.scala

Server Restructuring.md
tristantarrant: I don't want to bloat the package with a bunch of drivers, which the user would  want to update independently anyway. Better use META-INF/services/java.sql.Driver discovery

PR #3, opened 2017-04-10
Multimap design with mail discussion

karesti:
Design for https://issues.jboss.org/browse/ISPN-7780

### DIFF
--- a/CacheMultimap.adoc
+++ b/CacheMultimap.adoc
@@ -1,0 +1,79 @@
+= Native cache Mutimap
+
+The goal is to provide Infinispan Native CacheMultimap support. The proposal here is to start with the smallest interface implementation, in order to provide a proper API to the first multimap client, vert-x project.
+However, this API will evolve, and this first implementation must work on Embedded and Server mode.
+This first Multimap won't support duplicate values on the same key. 
+
+[source, java]
+.CacheMultimap.java
+----
+public interface CacheMultimap<K, V> {
+   boolean put(K key, V value); // <1> <2>
+
+   Collection<V> get(K key); // <3>
+
+   boolean remove(K key); //<4>
+
+   boolean remove(K key, V value);
+   
+   void remove(Predicate<V> p); //<5>
+}
+----
+<1> This method could return void as usual or a boolean. The first version of CacheMultimap won't allow duplicates. The boolean could indicate true if the value was been added and false if the value was not put.
+<2> "Put" indicates that the value is being somehow replaced on a key if the key already exists. An alternative could be to call it "add"
+<3> We might want to return a Set instead of Collection interface, because duplicates on same key won't be supported yet.
+<4> Calling this method "reset" has been suggested
+<5> vert-x implementation needs a way to remove values depending on a given Predicate. To achieve this, we need to provide an API or CacheSet<K> keySet(); method. 
+
+An alternative is to support only Async API instead of the standard sync API. 
+
+[source, java]
+.CacheMultimap.java
+----
+public interface CacheMultimap<K, V> {
+   CompletableFuture<Void> put(K key, V value);
+
+   CompletableFuture<Collection<V>> get(K key);
+
+   CompletableFuture<Boolean> remove(K key);
+
+   CompletableFuture<Boolean> remove(K key, V value);
+   
+   CompletableFuture<Void> remove(Predicate<V> p);   
+}
+----
+
+
+The underlying implementation will wrap a normal Cache. Some of the suggestions :
+
+== Core Module
+
+[source,java]
+----
+public interface EmbeddedCacheManager {
+ ...
+ 
+ <K, V> CacheMultimap<K, V> getCacheMultimap(String cacheName, boolean createIfAbsent);
+}
+
+----
+
+== Commons Module
+
+[source, java]
+----
+public class CacheMultimapFactory {
+
+   static <K, V> CacheMultimap<K, V> CacheMultimapFactory.get(Cache<K,Object> c) {
+      ...
+   }
+}
+----
+
+Apparently there is a problem with functional commands. They don't really work efficiently over Hot Rod (does get/replace in a loop). We would need to add some more handling in the protocol to allow for only partial replication of values and only 1 remote call.
+
+
+
+
+
+
--- a/CacheMultimap.adoc
+++ b/CacheMultimap.adoc
@@ -7,30 +7,29 @@ This first Multimap won't support duplicate values on the same key.
 [source, java]
 .CacheMultimap.java
 ----
-public interface CacheMultimap<K, V> {
-   boolean put(K key, V value); // <1> <2>
+public interface MultimapCache<K, V> {
+   void put(K key, V value); // <1>
 
-   Collection<V> get(K key); // <3>
+   Collection<V> get(K key); // <2>
 
-   boolean remove(K key); //<4>
+   boolean remove(K key); //<3>
 
    boolean remove(K key, V value);
    
-   void remove(Predicate<V> p); //<5>
+   void remove(Predicate<V> p); //<4>
 }
 ----
-<1> This method could return void as usual or a boolean. The first version of CacheMultimap won't allow duplicates. The boolean could indicate true if the value was been added and false if the value was not put.
-<2> "Put" indicates that the value is being somehow replaced on a key if the key already exists. An alternative could be to call it "add"
-<3> We might want to return a Set instead of Collection interface, because duplicates on same key won't be supported yet.
-<4> Calling this method "reset" has been suggested
-<5> vert-x implementation needs a way to remove values depending on a given Predicate. To achieve this, we need to provide an API or CacheSet<K> keySet(); method. 
+<1> "Put" indicates that the value is being somehow replaced on a key if the key already exists. An alternative could be to call it "add"
+<2> We might want to return a Set instead of Collection interface, because duplicates on same key won't be supported yet.
+<3> Calling this method "reset" has been suggested
+<4> vert-x implementation needs a way to remove values depending on a given Predicate. To achieve this, we need to provide an API or CacheSet<K> keySet(); method.
 
 An alternative is to support only Async API instead of the standard sync API. 
 
 [source, java]
-.CacheMultimap.java
+.MultimapCache.java
 ----
-public interface CacheMultimap<K, V> {
+public interface MultimapCache<K, V> {
    CompletableFuture<Void> put(K key, V value);
 
    CompletableFuture<Collection<V>> get(K key);
@@ -46,33 +45,52 @@ public interface CacheMultimap<K, V> {
 
 The underlying implementation will wrap a normal Cache. Some of the suggestions :
 
-== Core Module
 
-[source,java]
-----
-public interface EmbeddedCacheManager {
- ...
- 
- <K, V> CacheMultimap<K, V> getCacheMultimap(String cacheName, boolean createIfAbsent);
-}
-
-----
-
-== Commons Module
-
-[source, java]
-----
-public class CacheMultimapFactory {
-
-   static <K, V> CacheMultimap<K, V> CacheMultimapFactory.get(Cache<K,Object> c) {
-      ...
-   }
-}
-----
-
-Apparently there is a problem with functional commands. They don't really work efficiently over Hot Rod (does get/replace in a loop). We would need to add some more handling in the protocol to allow for only partial replication of values and only 1 remote call.
+Apparently there is a problem with functional commands.
+They don't really work efficiently over Hot Rod (does get/replace in a loop).
+We would need to add some more handling in the protocol to allow for only partial replication
+of values and only 1 remote call.
 
 
+== Embedded Multimap
+
+A new maven module is created.
+
+To create a multimap cache
+
+```java
+      EmbeddedCacheManager cm = ...
+      MultimapCacheManager multimapCacheManager = EmbeddedMultimapCacheManagerFactory.from(cm);
+      multimapCache = multimapCacheManager.get("test");
+
+      multimapCache.put("k", "v");
+
+```
+
+
+== Infinispan Server Multimap
+
+There are at least two options for the implementation
+
+=== Option A
+
+We define a new interface called RemoteMultimapCache that won't support methods with lambdas, but the most
+common and useful methods will be supported.
+
+We implement it using the existing OperationsFactory. We enrich the header or we add a flag so the server will know that this
+call is meant to be over a multimap. So the internal implementation will grab the cache manager and the cache, create the
+ EmbeddedMultimapCache and call the methods over that instance instead of over the regular cache methods.
+
+- + Code reuse
+
+- - Spaghetti code risk
+
+=== Option B
+As for Counters, Locks and any other module built on top of the core, we will implement new operations for hotrod that are specific
+for multimaps.
+
+- + Clear separation of modules
+- - Duplication of code risk
 
 
 


CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+ <K, V> CacheMultimap<K, V> getCacheMultimap(String cacheName, boolean createIfAbsent);
+}
+
+----
+
+== Commons Module
+
+[source, java]
+----
+public class CacheMultimapFactory {
```
rvansa: Why should the factory be in commons at all?

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+The underlying implementation will wrap a normal Cache. Some of the suggestions :
+
+== Core Module
+
+[source,java]
+----
+public interface EmbeddedCacheManager {
+ ...
+ 
+ <K, V> CacheMultimap<K, V> getCacheMultimap(String cacheName, boolean createIfAbsent);
```
rvansa: We were considering not adding the dependency into EmbeddedCacheManager at all, using `static EmbeddedCacheMultimapFactory.create(Cache<K, Object> cache)`

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+However, this API will evolve, and this first implementation must work on Embedded and Server mode.
+This first Multimap won't support duplicate values on the same key. 
+
+[source, java]
+.CacheMultimap.java
+----
+public interface CacheMultimap<K, V> {
+   boolean put(K key, V value); // <1> <2>
+
+   Collection<V> get(K key); // <3>
```
rvansa: Should this be a `Collection` or a `Set`?

Sanne: The original email thread mentioned that "this version" doesn't support repeated entities but I believe they meant that this is a temporary limitation?

If it's a temporary milestone only, it should be `Collection` so to allow getting rid of the limitation soon.

If it's not a temporary limitation and indeed this should return a `Set` .. maybe this whole API shouldn't be called `MultiMap` ? That would be very confusing...

karesti: vert-x needs a MultiMap that won't support duplicates on a key, but other existing multimap implementations allow duplicates. The question is : do we want and need duplicates ? As far as I remember @wburns considers we should be able to support duplicates. Honestly, I don't know if we should or shouldn't, but I guess we could handle both implementations in the future, so Collection fits better on the API.

CacheMultimap.adoc
karesti: This was my suggestion @rvansa, to put it in core, so I put it here from the e-mail.

CacheMultimap.adoc
karesti: This is what I understood on the discussion, but I had everything wrong maybe ? I understood as we need to provide the API for embedded and server, somehow the code, wrapper or factory should be in commons. I might be missing what the proposal really is... 😢

CacheMultimap.adoc
Sanne: > vert-x needs a MultiMap that won't support duplicates on a key

Are you sure of that? Why is it not called a "Map" then?

I have no opinion on what you should implement, but if you don't plan to support multiple values per key then it shouldn't be called "MultiMap".

rvansa: @Sanne duplicates in the collection, under one key you hold multiple items, so it is a multimap. The only question is whether there should be multiple equal items under the same key: `myKey -> [a, b]` vs. `myKey -> [a, a, b]`

CacheMultimap.adoc
rvansa: @karesti I think that CacheMultimap interface should be in the commons, since that one should be shared. However, the way how you obtain an instance of implementation may vary (you have to provide either embedded or remote cache).

CacheMultimap.adoc
Sanne: ah right! Sorry I was confused by that.

CacheMultimap.adoc
karesti: Ok ! so the internal factories and everything else should be in core then, right ?

rvansa: I would say that in a separate module, but core could work as well.

karesti: Ok 👍

CacheMultimap.adoc
tsegismont: > vert-x needs a MultiMap that won't support duplicates on a key

To be clear, Vert.x does not rely on the cluster manager implementation to avoid duplicate key/value pairs.

But it is true that for the event bus subscriptions use case, duplicates make no sense.

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+   boolean remove(K key); //<4>
+
+   boolean remove(K key, V value);
+   
+   void remove(Predicate<V> p); //<5>
+}
+----
+<1> This method could return void as usual or a boolean. The first version of CacheMultimap won't allow duplicates. The boolean could indicate true if the value was been added and false if the value was not put.
+<2> "Put" indicates that the value is being somehow replaced on a key if the key already exists. An alternative could be to call it "add"
```
tsegismont: Put is just fine IMHO

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+<3> We might want to return a Set instead of Collection interface, because duplicates on same key won't be supported yet.
```
tsegismont: I think Collection is better since, if you support duplicates later, the collection could behave like a list or a set.

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+<4> Calling this method "reset" has been suggested
```
tsegismont: IMO, the usual cache/map method names (get/put/remove) are fine.

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+<5> vert-x implementation needs a way to remove values depending on a given Predicate. To achieve this, we need to provide an API or CacheSet<K> keySet(); method.
```
tsegismont: With existing caches (dist, repl) is there a way to modify the data in a distributed way?

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+   void remove(Predicate<V> p); //<5>
+}
+----
+<1> This method could return void as usual or a boolean. The first version of CacheMultimap won't allow duplicates. The boolean could indicate true if the value was been added and false if the value was not put.
+<2> "Put" indicates that the value is being somehow replaced on a key if the key already exists. An alternative could be to call it "add"
+<3> We might want to return a Set instead of Collection interface, because duplicates on same key won't be supported yet.
+<4> Calling this method "reset" has been suggested
+<5> vert-x implementation needs a way to remove values depending on a given Predicate. To achieve this, we need to provide an API or CacheSet<K> keySet(); method. 
+
+An alternative is to support only Async API instead of the standard sync API.
```
tsegismont: Vert.x would most probably use the async version. But shouldn't the API look like the usual cache hierarchy (Cache extends BasicCache, which  extends AsyncCache) ?

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
+= Native cache Mutimap
+
+The goal is to provide Infinispan Native CacheMultimap support. The proposal here is to start with the smallest interface implementation, in order to provide a proper API to the first multimap client, vert-x project.
+However, this API will evolve, and this first implementation must work on Embedded and Server mode.
+This first Multimap won't support duplicate values on the same key. 
+
+[source, java]
+.CacheMultimap.java
+----
+public interface CacheMultimap<K, V> {
```
tsegismont: Shouldn't the name be Multicache? Does ISPN have any other construct named after map?

CacheMultimap.adoc
rvansa: Do you need a method to remove whole key-collection pair if the key/collection matches some criteria (therefore it would be `removeAll(BiPredicate<K, Collection<V>>`), or just remove values from single pair (`remove(K key, Predicate<V>)`)?

CacheMultimap.adoc
rvansa: Just because other features with 8 years of history, before the async models emerged, carry this burden? When designing on a green field, lets make the API exactly the way we want to use it.

tsegismont:
Just fine with me. Not sure everyone expects an async API only though.

Anyway, ISPN team call :)

2017-04-11 10:17 GMT+02:00 Radim Vansa <notifications@github.com>:

tsegismont:
Just the latter, remove(K key, Predicate<V>)

In the subscriptions multimap, we store <String, ClusteNodeInfo> pairs, the
key being an event bus address, the value a node having a handler
registered for this address.

2017-04-11 10:13 GMT+02:00 Radim Vansa <notifications@github.com>:

CacheMultimap.adoc
tsegismont: BTW, this method is not needed in the Vert.x cluster manager. But it could be useful to others I guess.

CacheMultimap.adoc
karesti: Galder thoughts

CompletableFuture is good.

In hindsight, I would have maybe chosen java.util.concurrent.CompletionStage since it's more flexible (interface vs class), and doesn't bring in java.util.concurrent.Future which contains blocking methods.
CompletableFuture/CompletionStage works for single returns. The bigger problem is when you want multiple returns asynchronously. Here a can of worms opens up, e.g. do you push the results? do you pull the results?

For the functional map API, we experimented with a pull model using Traversable. A push model is harder to implement and  non-trivial, and there you're getting into Rx territory.

karesti:
In order to move forward the design, this multimap should be available for server mode. 

I guess, we should create a new interface giving access to the fonctionality in server mode. 
And create a simple example that demo's the multimap.

``` java
public interface RemoteCacheMultimap<K, V> extends CacheMultimap<K, V> {
...
}

```

tristantarrant:
No need to have a separate interface for remote, unless there are additional methods (which might be the case if we need access to versions)

karesti:
@tristantarrant  get method would be returning a Collection, and there are other new methods too. IMHO a new interface for multimap support is the cleanest way to go, even for remote API.

rvansa:
@karesti I think that @tristantarrant was saying that the interface for embedded and c/s should be the same. Unless we want conditional execution, which is in case of Hot Rod best done using the [versioned extensions](https://github.com/infinispan/infinispan/blob/master/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCache.java#L207). Please also checkout https://docs.google.com/document/d/1y-eLiZ6657qevHYlXXPI9VeY0sN2XIg-ykea9ZUVo4Q/edit#heading=h.tenj92huu8a1 for current deficiencies in the API.
If we want to extend the proposed multimap API with versions, I think that the crucial decision is whether we want to version whole collection, each element in the collection or both.

karesti:
@rvansa ok ! my bad ! thank you for the document. I'm going to read it now

rvansa:
How is the multimap supposed to behave in transactional cache with write-skew-check enabled? Should the WSC fire when you only call `put()s` or `remove(key, value)` (with a different value)? What about remove with predicate - will that count as a read?

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+   boolean put(K key, V value); // <1> <2>
+
+   Collection<V> get(K key); // <3>
+
+   boolean remove(K key); //<4>
+
+   boolean remove(K key, V value);
+   
+   void remove(Predicate<V> p); //<5>
```
rvansa: this could return number of removed elements (since remove tells you if that removes anything). `Collection<V> getAndRemove(Predicate<V> p)` would be useful, too, returning the removed entities.
And if I am not mistaken the correct generics for this should be `Predicate<? super V>`.

karesti: O for the generics and the return value on int ! For the proposed method return with deleted values, why not, but I prefer concentrating on the "smallest" fancy API 💃

karesti:
@rvansa Concerning transactions with write-skew, as the underlying first implementation will be a cache, the way this would behave is the same as having a cache today with a String and a List with values. We might want to improve the flexibility of the multimap in this area later. What do you think ?

rvansa:
@karesti That's a vague definition. Do you mean that the result of an operation is the same as reading the previous List and replacing that with a different List? That means that when executing e.g. two put()s with different values one of them will fail with WSC. I don't say that this is inevitably wrong; but we should think about it.

Remember that this is the behaviour == API, too, API's not just the signature of methods & parameters. You can't "improve" that later unless adding knobs.

And please include this info in Javadoc (preferably stating that in design docs first, so you won't forget). You'll have to cover that in the testsuite, too.

CacheMultimap.adoc
karesti: I'm keeping in core

CacheMultimap.adoc
karesti: @tsegismont I'm doing the implementation and I came to the same conclusion. It's better to swap the names, so calling it MultimapCache rather than CacheMultimap.

CacheMultimap.adoc
karesti: @rvansa After some time, I've came up to the personal conclusion that would be better to keep the multimap cache creation via RemoteCacheManager and EmbeddedCacheManager.
I know for Counters for example, we are doing things like EmbeddedCounterManagerFactory so that we can access to the CounterManager and create counter from there.

From an usability, api discovery and learning point of view for users, I think this is not a good idea. Having a common entry point for API's used by user's is important.

rvansa: @karesti If the implementation will be in core module, regular method on `EmbeddedCacheManager` would be appropriate - the whole point of the reverse approach is making core module independent on the implementations of various add-ons like counters, multimaps, locks and other builiding blocks on top of key-value storage.

In the end, it's @tristantarrant 's call to put ubermodule and loose module system on the scales.

karesti: @rvansa @tristantarrant som for multimap or locks, another maven module is where we want to go, right ? like the one done for counters

CacheMultimap.adoc
tsegismont: What we need actually is sth like `remove(Predicate<V>)`
The Vert.x subscriptions multimap maps event bus addresses to a collection of server ids (which nodes have a consumer registered for an address).
When a server leaves the cluster we have to remove all values which match a predicate. So `remove(Predicate<V>)`

CacheMultimap.adoc
wburns: I think dropping the word Map as @tsegismont  suggested is probably a good idea (ie. MultiCache). Especially if we end up changing the API to not be an exact mirror of Map as we have currently.

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+The goal is to provide Infinispan Native CacheMultimap support. The proposal here is to start with the smallest interface implementation, in order to provide a proper API to the first multimap client, vert-x project.
+However, this API will evolve, and this first implementation must work on Embedded and Server mode.
+This first Multimap won't support duplicate values on the same key. 
+
+[source, java]
+.CacheMultimap.java
+----
+public interface CacheMultimap<K, V> {
+   boolean put(K key, V value); // <1> <2>
```
wburns: Technically if we allow duplicates the return value is worthless.

CacheMultimap.adoc
wburns: We should also document what this returns if there are no values for the given key. Do we ever return a null or do we just return an empty collection?

CacheMultimap.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+   Collection<V> get(K key); // <3>
+
+   boolean remove(K key); //<4>
```
wburns: Do we want to fall into the same trap as `Map` vs `JCache`? I wonder if we want to have `getAndRemove` and `remove` combinations. Same with `getAndPut` and `put` etc.

karesti: @wburns concerning the Collection, indeed, I realized while I was developing it. It should return an empty collection when the value does not exist, not null.  If we added a method like containsKey(), the result will be false.
When a key has an empty collection - after a remove for example - the key will be removed.

karesti: @wburns concerning the JCache and Map, I don't understand the trap... we can add those methods after if we want them, can't we ? 😕

CacheMultimap.adoc
karesti: you are right. I think void will be better

CacheMultimap.adoc
wburns: the problem is that if the API defines having a return type then we need to actually send back the return (ie. getAndPut). But there may be cases where the user only wants to put the value and doesn't care about the return value (ie. put). The JCache API defines put as being void for example.
We have a way to work around this despite inheriting from Map with our `Flag.SKIP_REMOTE_LOOKUP` but no one knows to use this.

rvansa: @wburns `Flag.IGNORE_RETURN_VALUE` - `SKIP_REMOTE_LOOKUP` is one of the hazardous options (saying 'if a node needs a value and does not have it locally, use null, but don't try to contact other nodes').

wburns: @rvansa yes my point exactly

ben-manes: The choice doesn't have to be to suffer with a Map biased interface or be hostile towards the Collections Framework like JCache is. You could provide an idiomatic API with no needless return types and an asMap() view as an escape hatch. That has worked well in Guava/Caffeine and avoids a single large, complex Interface that isn't idiomatic with the rest of Java.

karesti: @ben-manes @wburns @rvansa what if we do the "put" as void and expose a method called asCache Cache<K,Collection<V>> ?

rvansa: @karesti And what are you trying to solve? "void put" or "void remove" just don't force the code to load the collection into memory (though in practice you'll have to, as you can't send a function to the cache store), and not force to transfer any data over the wire.
So you can expose:
1. void put(), void remove()
2. boolean putAndReturnTrueIfSomeDataWasPresent(), boolean removeAndReturnTrueIfSomeEntryWasPresent()
3. the same as 2 but returning integers with collection sizes
4. Collection<V> putAndReturnAllPreviousEntries(), Collection<V> removeAndReturnThoseRemovedEntries();
5. combination of 1. and 2./3. and 4.

Since you have to load the data anyway and the cost of marshalling booleans/integers is neglible, I personally like 3. the most (4. could eventually get in *in addition to 3.* but not necessarily in the first draft). But I don't mind any variant if there's other option than 4.

danberindei: @rvansa -1 for option 3. A boolean return value tells the user whether a `key->value` mapping was added or not (because it already existed). I'm not sure how useful that would be, but the prior collection size seems pretty useless.

@karesti I wouldn't expose the full Cache API. Users that need more complex processing of the values, e.g. full-blown `compute()`, should use `Cache<K, Collection<V>>` directly.

rvansa: @karesti looking again on the return value of `Map.compute()`, if you want to utilize this for the multimap, you need to use `void put()` and `void remove()`, since the `remappingFunction` can't return one value and store another - `compute()` would return the new value and seemingly you need to use `cache.withFlags(Flag.IGNORE_RETURN_VALUE).compute(key, addEntryToCollectionFunction)`.
Can you see something else that could make 2. or 3. work?

karesti: @rvansa @danberindei I'm using eval from functional interface now 💃 We should consider removing the "Experimental" annotation on the functional map feature, because this is being used in counters too

rvansa: @karesti `@Experimental` means that the API could change in non-major versions, so for internal use it's perfectly ok. It's not like `@ThisCouldFormatYourDrive` :) But I agree that we could remove those annotations in like 9.2.

CacheMultimap.adoc
tristantarrant: +1 to MultiCache

karesti:
I've done a PR with a tech preview of the discussion in this thread

https://github.com/infinispan/infinispan/pull/5193

CacheMultimap.adoc
```diff
@@ -46,33 +45,52 @@ public interface CacheMultimap<K, V> {
...
-   static <K, V> CacheMultimap<K, V> CacheMultimapFactory.get(Cache<K,Object> c) {
-      ...
-   }
-}
-----
+To create a multimap cache
+
+```java
+      EmbeddedCacheManager cm = ...
+      MultimapCacheManager multimapCacheManager = EmbeddedMultimapCacheManagerFactory.from(cm);
```
rvansa: Do we need the `MultimapCacheManager`? Why don't you just pass the underlying cache to the factory?

CacheMultimap.adoc
```diff
@@ -46,33 +45,52 @@ public interface CacheMultimap<K, V> {
...
+call is meant to be over a multimap. So the internal implementation will grab the cache manager and the cache, create the
+ EmbeddedMultimapCache and call the methods over that instance instead of over the regular cache methods.
+
+- + Code reuse
 
-Apparently there is a problem with functional commands. They don't really work efficiently over Hot Rod (does get/replace in a loop). We would need to add some more handling in the protocol to allow for only partial replication of values and only 1 remote call.
+- - Spaghetti code risk
 
+=== Option B
+As for Counters, Locks and any other module built on top of the core, we will implement new operations for hotrod that are specific
```
rvansa: +1

CacheMultimap.adoc
karesti: @tristantarrant prefer this from doing it the other way. This is how counters are done too, makes the api more coherent.

rvansa: Hmm, ok. The difference is that there is single cache providing services to all counters, while there's a 1:1 relation between caches and multimaps. Will you prefix the cache names for multimap operations, or do you plan to openly admit that the cache can be accessed using the same name?

karesti: @rvansa I was thinking that we might prefix cache names and config with multimap. I didn't think too much about it yet because this is the first version and I'm planning to mark it as experimental.

tristantarrant: As we don't have cache namespaces, or a naming convention about this, we should probably just go for a shared namespace, and potentially accept access through the traditional cache API

CacheMultimap.adoc
```diff
@@ -0,0 +1,97 @@
...
+- + Code reuse
+
+- - Spaghetti code risk
+
+=== Option B
+As for Counters, Locks and any other module built on top of the core, we will implement new operations for hotrod that are specific
+for multimaps.
+
+- + Clear separation of modules
+- - Duplication of code risk
```
tristantarrant: Separation wins over duplication

CacheMultimap.adoc
karesti: @tristantarrant yes, I I said before, this would be the best right now. The PR for the first version is ready !

CacheMultimap.adoc
karesti: Concerning hotrod, I already did the hotrod part for multimap. But I found particulary difficult to separate into a different module without duplicating so much code. Gustavo's encoding PR is concerned on this, because I had to implement the encoding on functional maps (which was not supported) in order to make hot rod work.

PR #4, opened 2017-04-18
Clustered Cache Configuration and State Design

tristantarrant:


Clustered-cache-configuration-state.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+
+* Add a ___state cache handled by the InternalCacheRegistry. This cache will contain both cache configurations (stored
+as strings using the Infinispan XML schema) and a list of running caches (do we need to store more state ?).
+* Introduce a new ClusteredCacheContainer API (see below)
+* Cache configuration and lifecycle performed via the old DefaultCacheManager API will continue to behave as it does
+currently (i.e. each node is independent)
+* Allow configuration Attributes to be either Global or Local.
+* Implement a variant of equals (equalsIgnoreLocal ?) for Configuration objects to validate congruency ignoring local
+attributes.
+* Support the +start+ configuration attribute for caches (which can be either +LAZY+ or +EAGER+) so that EAGER caches
```
tristantarrant: EAGER/LAZY comes from WildFly. In server it has always been ignored, since we never had a way to start a cache across the cluster when accessed through the remote endpoint. In embedded mode, it was never a thing

tristantarrant:
@rvansa actually we don't need to create a new interface: this could be EmbeddedCacheManager itself. Whatever we do, I do not want to create more methods (defineClusteredConfiguration)

PR #5, opened 2017-04-28
Proper command retry handling

rvansa:
Design for https://github.com/infinispan/infinispan/pull/5064

### DIFF
--- a/Proper-command-retry-handling.md
+++ b/Proper-command-retry-handling.md
@@ -1,0 +1,35 @@
+This document summarizes changes introduced in [ISPN-7590](https://github.com/infinispan/infinispan/pull/5064).
+
+# Goals
+* Command should never be applied to an entry twice. 
+* If a command does not ignore previous value (on backup), it must be applied on the backup to the same value as on primary.
+* The return value of a command should be reflect the change regardless of retries (e.g. replace will return old value even if it is executed multiple times)
+* Events should have correct type (created/modified)
+* Events should be fired on each node only once
+
+# Non-goals
+* Correct previous value in events - events won't be reliable anyway as some commands don't load previous value from persistence/remote node etc.
+* Solve [ISPN-3918](https://issues.jboss.org/browse/ISPN-3918) - although this was amongst the goals initially, the fix would have negative performance impact or would be out of scope (flag-affected behaviour).
+* Fix anything related to functional events
+* Deal with delta-awares more than what is required to pass current testsuite
+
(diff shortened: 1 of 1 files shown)


Proper-command-retry-handling.md
```diff
@@ -0,0 +1,35 @@
...
+  * In async modes we can't do any invalidations because originator does not know when the command finishes; we have to rely on expiration. However, we have to store the invocation record even if the entry is committed with `ctx.isOriginLocal()`. Regrettably this forces us to store the information about command being executed in sync/async mode in the command itself.
+* Commands that require the same previous value on backup as on primary implement `StrictOrderingCommand`. At this point these are ReadWrite* functional commands. When a strict command is replicated to backup, primary adds command invocation id of previous command (or null if there's no previous record). The backup checks its last record against this last invocation id and applies the command only in case of match. In other case it retrieves full entry including metadata with invocation records from the primary owner, and stores this value (potentially applying the command it it was not applied - this depends on classical/triangle routing algorithms).
```
galderz: Seems like this would get rid of `ReadWriteKeyValueCommand` shipping `prevValue` and `prevMetadata` around? Your approach would appear to make the command more lightweight, and instead use the command id of previous command affecting that key/value pair.

However, if no `prev*` is shipped, you'd have to get `prev*` data from the `InvocationManager` on demand, right?

It'd be good to explain this a bit further, since from my POV, one of the key reasons why functional API is not yet graduated to not being experimental is due to the cost required to deal with retries. As I've explained before, if you have a List of 1000 elements and you want to pass a lambda around to add an element in the middle of the list, having to ship around the previous value defeats the purpose of functional cmds.

galderz:
@rvansa Interesting approach :)

Proper-command-retry-handling.md
rvansa: Yes, `prev*` is a temporary hack, and there's no point in using functional writes if you ship previous value.

I don't really follow your question: the `prev*` data should be on the backup owner (where you apply the command) most of the time. There are some cases when these may be missing, and in that case we need to retrieve the value from the primary (through `RpcManager`, though, not `InvocationManager`).

galderz: Ok, that answers my question. I wanted to know in more detail how the retrieval of values not present locally would work.

rvansa:
Closing for now - in Como we've discussed alternative approach, because the current one does not fix continuous query.

PR #6, opened 2017-05-05
Design document for Hot Rod Transactions

pruivo:
* Lightweight & Synchronization enlistment only

### DIFF
--- a/Hot-Rod-Transactions-NonXA.asciidoc
+++ b/Hot-Rod-Transactions-NonXA.asciidoc
@@ -1,0 +1,313 @@
+Lightweight transactions for Hot Rod
+====================================
+
+*JIRA*: https://issues.jboss.org/browse/ISPN-7807[ISPN-7807]
+
+== Introduction
+
+This document is a design document for lightweight Hot Rod transaction and Synchronization enlistment. Full XA support will be described in another document.
+
+The transaction state is kept in the client during the transaction execution and the server is only aware of the transaction when the client wants to commit.
+
+== Changes to existing code
+
+Since a default `TransactionManager` is needed for the client, a base implementation is extracted from the `EmbeddedTransactionManager` to commons package.
+The `Transaction`'s implementation (and it dependencies) will be moved to commons package.
+
+== Limitation and Requirements
+
+* In the client, the transaction will be enlisted as `Synchronization`.
+XA support will be added later in another document.
+
+* The server will use `REPEATABLE_READ` isolation.
+
+* The server will use the `EmbeddedTransactionManager`.
+It needs to invoke the prepare and commit/rollback phase individually.
+
+* [.line-through]#Recovery needs to be enabled in the server.
+It prevents the transaction to be automatically rolled-back when the originated leaves the cluster.#
+
+* Any locking is supported.
+
+== Transaction Enlistment
+
+The remote cache is registered in the `TransactionManager` as `Synchronization`.
+
+A global `SynchronizationAdapter` will be implemented per `RemoteCacheManager`.
+This means that if multiple remote caches are involved in a transaction, only a single `Synchronization` is registered in the transaction.
+It main goal is to translate and forward the requests from `TransactionManager` to the remote caches.
+
+When multiple remote caches are involved, the prepare phase is handled synchronously and orderly.
+The main reasons for this are:
+
+* *safety*: avoids any deadlocks between remote caches since the remote caches are prepared at the same order.
+* *optimization*: if a remote cache fails to prepare, the remaining remote caches are not contacted.
+
+The second phase (commit/rollback) can be done in parallel in all the remote caches, but for the first implementation, it will be done sequentially.
+
+== Transaction's Isolation Level
+
+As said, only `REPEATABLE_READ` will be supported. The reasons are:
+
+* It is the strongest guarantee supported by the server,
+
+* It offers a better performance by skipping interaction with the server at each requests,
+
+* In the future, `READ_COMMITTED` will be removed.
+
+== Description
+
+Each remote-cache will have a transaction context associated to a specific transaction.
+Each key read or write will be stored in this context.
+Also, metadata needed for transaction validation is stored in the context.
+
+At the first operation, the remote cache registers itself as a participant in the `SynchronizationAdapter`.
+
+=== Transaction execution
+
+Read operations is executed as following:
+
+* If the key exists in transaction context, return its value;
+* Otherwise, fetch the value and metadata from the server and store it in the transaction context.
+
+Write operations is executed as following:
+
+* If the key doesn't exists in the transaction context and it is required (conditional value or needs to return previous value), fetches value and metadata from the server and store it in the transaction context;
+* Applies the operation based on the value in the transaction context and updates its value.
+
+NOTE: The value and metadata can be fetch from any node in the cluster.
+
+=== Transaction commit
+
+A `Transaction` is committed in 2 phases: `beforeCompletion()` and `afterCompletion()`.
+
+In `beforeCompletion()`, the client collects the written keys and sends them to the server (any member in cluster).
+The following information is sent:
+
+* XID - this is the transaction id;
+* For each key, it sends:
+** The key;
+** The `ReadState`;
+** The version read (if any);
+** The new value;
+** The metadata;
+
+[NOTE]
+======
+The `ReadState` represents the following information:
+
+* `NOT_READ`: the key wasn't read before the write;
+* `NON_EXISTING`: the key was read before write and it didn't exist;
+* `READ`: The key was read before write and it exists (the versionr read is required)
+======
+
+In the server side, when the request is received, it replays the transaction.
+A new transaction is created and set its XID based on the XID received (to avoid any XID translation).
+
+For each key received, the server performs the following:
+
+* It checks the `ReadState` value:
+** `NOT_READ`: it doesn't fetch the previous value.
+** `NON_EXISTING`: it checks if the key doesn't exist.
+If the key exists, the transaction is rolled-back.
+** `READ`: it fetches the value and metadata.
+Then it compares the version received with the version stored in the metadata.
+If they don't match, the transaction is rolled-back.
+* Finally, it performs the write.
+
+When all keys are handled, the server invokes `EmbeddedTransaction.runPrepare()`.
+The outcome is sending back to the client.
+
+The client receives the outcome and marks the transaction to rollback if the server failed to prepare.
+
+Finally, in `afterCompletion()`, the client only sends the commit or rollback request:
+
+* XID
+* `COMMIT` or `ROLLBACK` request.
+
+In the server, if the transaction was executed locally, it invokes the `EmbeddedTransaction.runCommit(`).
+Otherwise, use the `RecoveryManager` to commit or rollback the transaction.
+
+NOTE: Not sure if the `RecoveryManager` can be used or if I need to create a Global Transaction Table.
+
+== Hot Rod Protocol Details
+
+=== Operation Codes
+
+4 new operation codes:
+
+|===
+|Operation Code | Response Operation Code |Operation
+
+|0x3B
+|0x3C
+|Prepare Request
+
+|0x3D
+|0x3E
+|Commit Request
+
+|0x3F
+|0x40
+|Rollback Request
+
+|===
+
+=== Messages description
+
+The transaction id (XID) is present on the request header.
+However, it will be changed to support variable size XID
+
+.XID
+[cols="<.^,<.^,<"]
+|===
+|Field Name |Size |Value
+
+|Format ID
+|long
+| The XID format.
+
+|Length of Global Transaction id
+|byte
+|The length of global transaction id byte array. It max value is `64`.
+
+|Global Transaction Id
+|byte array
+|The global transaction id.
+
+|Length of Branch Qualifier
+|byte
+|The length of branch qualifier byte array. It max value is `64`.
+
+|Branch Qualifier
+|byte array
+|The branch qualifier.
+
+|===
+
+
+.Prepare Request (operation code `0x3B`)
+[cols="<.^,<.^,<"]
+|===
+|Field Name |Size |Value
+
+|Header
+|variable
+|Request header
+
+|OnePhaseCommit
+|byte
+|When it is set to `1`, the server will use one-phase-commit if available (XA only)
+
+|Number of keys
+|vInt
+|The number of keys
+
+3+^|For each key
+
+| Key Length
+| vInt
+| Length of key.
+Note that the size of a vInt can be up to 5 bytes which in theory can produce bigger numbers than `Integer.MAX_VALUE`.
+However, Java cannot create a single array that’s bigger than `Integer.MAX_VALUE`, hence the protocol is limiting vInt array lengths to `Integer.MAX_VALUE`.
+
+|Key
+|byte array
+|Byte array containing the key
+
+|Version Read
+|long
+|The version read. `0` if `ReadState` is `NOT_READ` or `NON_EXISTING`.
+
+
+|Control Byte
+|Byte
+|The first 2 bits are the `ReadState`: +
++0x01+ = `NOT_READ` +
++0x02+ = `NON_EXISTING` +
++0x03+ = `READ` +
+The 3rd bit represents the remove or put operation.
+If `1`, the operation is a remove and the next data is not sent.
+If `0`, the operation is a put.
+
+|Lifespan
+|vInt
+|Number of seconds that a entry during which the entry is allowed to life.
+If number of seconds is bigger than 30 days, this number of seconds is treated as UNIX time and so, represents the number of seconds since 1/1/1970.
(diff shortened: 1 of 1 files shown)


Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+* Any locking is supported.
+
+== Transaction Enlistment
+
+The remote cache is registered in the `TransactionManager` as `Synchronization`.
+
+A global `SynchronizationAdapter` will be implemented.
+It main goal is to register itself and it will map a transaction to the remote caches involved.
+
+If multiple remote caches are involved in a single transaction, it commit will happen orderly between the remote caches.
```
rvansa: I've failed to understand this: will you invoke synchronous prepare for each cache in some (alphabetic?) order, and then synchronous commit? Or will you start prepare for each cache, wait until all caches prepare, and then start committing all caches & wait until all commit? I definitely vote for 2), though it doesn't work that way in embedded mode right now (yet).

I'm also unsure if there'll be single Synchronization registered for all caches, or one for each remote cache.

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+
+For each key received, the server performs the following:
+
+* It checks the `ReadState` value:
+** `NOT_READ`: it doesn't fetch the previous value.
+** `NON_EXISTING`: it checks if the key doesn't exist.
+If the key exists, the transaction is rolled-back.
+** `READ`: it fetches the value and metadata.
+Then it compares the version received with the version stored in the metadata.
+If they don't match, the transaction is rolled-back.
```
rvansa: Will the client get WriteSkewException?

rvansa: So the contacted server (originator) will need to fetch the data from remote node? Wouldn't it be better to prepare the data on each owner from client?

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+* Finally, it performs the write.
+
+When all keys are handled, the server invokes `EmbeddedTransaction.runPrepare()`.
+The outcome is sending back to the client.
+
+The client receives the outcome and marks the transaction to rollback if the server failed to prepare.
+
+Finally, in `afterCompletion()`, the client only sends the commit or rollback request:
```
rvansa: How is the timeout set (server has to rollback tx if client does not send commit/rollback)?

If the server crashes, will client retry the prepare on another server, or will it rollback immediately?

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+| Length of key.
+Note that the size of a vInt can be up to 5 bytes which in theory can produce bigger numbers than `Integer.MAX_VALUE`.
+However, Java cannot create a single array that’s bigger than `Integer.MAX_VALUE`, hence the protocol is limiting vInt array lengths to `Integer.MAX_VALUE`.
+
+|Key
+|byte array
+|Byte array containing the key
+
+|Version Read
+|vInt
```
rvansa: shouldn't this be vLong?

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+
+This solution covers the case where the client sends the prepare and the server crashes without sending the reply.
+
+=== Topology change during transaction second phase
+
+In this scenario, the client has the outcome of the prepare phase and it is free to retry the commit or rollback request in whatever node it wants.
+
+=== Client crashes/disconnects
+
+After a prepare phase, the locks are held until the client sends the commit or rollback request.
```
rvansa: Ouch, why not a rollback after timeout?

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: the first one for safety reasons. If we try to prepare all caches at the same time, we could have some deadlock between them. The second phase (commit/rollback) can be done in parallel but I'm not going to do it in the first implementation (or until I improve my knowledge of Hot Rod protocol)

it will be a single Synchronization per RemoteCacheManager (and all its caches). This is one of the changes required for embedded mode as well.

I'll rewrite this part :)

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: > Will the client get WriteSkewException?

No, it will only get the XA return/error codes. It is the only information it needs to reply to the TransactionManager.


> So the contacted server (originator) will need to fetch the data from remote node? Wouldn't it be better to prepare the data on each owner from client?

I'm not ready for this kind of optimization. you will end up in `N` local transaction in the server all with the same XID. I think it is too complex for the initial implementation.

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: nice catch :+1:

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: this is a tricky subject. You don't know when the commit will arrive and it is kind of bad of preparing the transaction and when the commit arrives, the transaction no longer exists.

ryanemerson: Could we not make this configurable? Locks being held indefinitely due to a client crash seems like a very bad idea, especially when multiple clients are accessing the cache.

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+* In the client, the transaction will be enlisted as `Synchronization`.
+XA support will be added later in another document.
+
+* The server will use `REPEATABLE_READ` isolation.
+
+* The server will use the `EmbeddedTransactionManager`.
+It needs to invoke the prepare and commit/rollback phase individually.
+
+* Recovery needs to be enabled in the server.
+It prevents the transaction to be automatically rolled-back when the originated leaves the cluster.
```
danberindei: Recovery requires manual intervention to clean up locks and a lot of other baggage (i.e. XIDs). I'd rather add a flag to client transactions so that the TransactionTable cleanup doesn't remove them when the originating node fails.

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+
+== Transaction's Isolation Level
+
+As said, only `REPEATABLE_READ` will be supported. The reasons are:
+
+* It is the strongest guarantee supported by the server,
+
+* It offers a better performance by skipping interaction with the server at each requests,
+
+* In the future, `READ_COMMITTED` will be removed.
```
danberindei: *may* be removed :)

Hot-Rod-Transactions-NonXA.asciidoc
danberindei: +1 to contact only one server per transaction for now.

Hot-Rod-Transactions-NonXA.asciidoc
danberindei: If we add a timeout, then it definitely needs to be configurable. Most transaction managers have a timeout after which they roll back inactive transactions, so this shouldn't be surprising to the users - we just need to make the exception meaningful.

pruivo: But I don't think that the transaction manager's timeout creates inconsistencies. If the transaction takes too long to run, it is safe to rollback. 
In our case, the transaction is prepared and it is bad the transaction manager requires the commit and we reply that "the transaction was aborted, sorry"

pruivo: BTW, I don't see any other alternative. But I don't like it. 
So, for an opinion, do you want a timeout configured per-transaction or common to all transactions?

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+The `Transaction`'s implementation (and it dependencies) will be moved to commons package.
+
+== Limitation and Requirements
+
+* In the client, the transaction will be enlisted as `Synchronization`.
+XA support will be added later in another document.
+
+* The server will use `REPEATABLE_READ` isolation.
+
+* The server will use the `EmbeddedTransactionManager`.
```
pruivo: I have no idea what compatibility mode is. can you explain? 
But I don't think that it supports preparing transactions and then suspend it.

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: When I started writing the doc, I was convinced that Recovery would fit nicely, but now I have my doubts :) 
I'll remove it :)

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: probably not this month (May) :)

Sanne:
Remember to update / reject related issues such as https://issues.jboss.org/browse/ISPN-375

Hot-Rod-Transactions-NonXA.asciidoc
Sanne: Remember that a timeout in the Transaction Manager (running on a client) is not the same thing as the transactions to automatically vanish from the resource (the "database"?). The TM might live on a client which is down, and need to perform a controlled recovery at a later time: not sure what happens if it can't possibly proceed at that point as the database decided to remove the locks it was owning.

danberindei: @pruivo I was thinking a single timeout for all the transactions. But now I see Narayana allows configuring the timeout [per transaction](http://narayana.io/docs/project/index.html#d0e2032), so maybe we need to support that too...

@Sanne We're not talking about supporting recovery on the client for now.

Hot-Rod-Transactions-NonXA.asciidoc
danberindei: http://infinispan.org/docs/stable/user_guide/user_guide.html#embedded_remote_compatibility
http://docs.oracle.com/javaee/6/api/javax/transaction/TransactionManager.html#suspend()

Hot-Rod-Transactions-NonXA.asciidoc
Sanne: @danberindei I know, I'm not suggesting you do implement it.

-  what I mentioned is not "on the client"
- it's just an example to highlight that what happens with "within database" stuff is not conceptually relates with what a Transaction Manager does on a client.

Just saying I wouldn't make any such design decision "because it's how a TM does things".. that's a wrong premise.

That being the wrong reason doesn't make it wrong though: I agree we'll need some form of cleanup.

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: compatibility mode won't affect how transactions runs in the client. 
the `suspend()` doesn't help either because the `TransactionManager` has no way to run the prepare phase only.

Hot-Rod-Transactions-NonXA.asciidoc
rvansa: You're right that simply running the prepare in parallel would lead to deadlocks. If we want to avoid the roundtrip for each cache, we have to lock keys in order according to <cache-name, key>, and that would mean having a multi-cache prepare command (on the wire, not in the interceptor stack) - and that would be a huge change.

However, on the level of HotRod, you could send the modifications for all caches in one go and let the server handle the details (if it's going to do prepare sequentially, as in the initial implementation, or if it adopts smarter strategy). Even with a sequential impl on server you'll spare some roundtrips to the server.

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,313 @@
+Lightweight transactions for Hot Rod
+====================================
+
+*JIRA*: https://issues.jboss.org/browse/ISPN-7807[ISPN-7807]
+
+== Introduction
+
+This document is a design document for lightweight Hot Rod transaction and Synchronization enlistment. Full XA support will be described in another document.
```
slaskawi: perhaps it would be good to mention `lightweight optimistic`?

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+* In the client, the transaction will be enlisted as `Synchronization`.
+XA support will be added later in another document.
+
+* The server will use `REPEATABLE_READ` isolation.
+
+* The server will use the `EmbeddedTransactionManager`.
+It needs to invoke the prepare and commit/rollback phase individually.
+
+* [.line-through]#Recovery needs to be enabled in the server.
```
slaskawi: I assume you mean we should enable it by default?

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+NOTE: Not sure if the `RecoveryManager` can be used or if I need to create a Global Transaction Table.
+
+== Hot Rod Protocol Details
+
+=== Operation Codes
+
+4 new operation codes:
+
+|===
+|Operation Code | Response Operation Code |Operation
```
slaskawi: Just a silly and naive idea. @tristantarrant mentioned that he used server tasks for implementing admin operations. What if we did the same trick here? Just create a script and send it to the server? This way we wouldn't need additional operations which needs to be implemented in *ALL* clients.

If this doesn't make sense, just ignore my comments. I'm thinking out laud...

galderz: I agree, it should be considered whether the commands for tx caches could be implemented using server tasks.

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+|0x3F
+|0x40
+|Rollback Request
+
+|===
+
+=== Messages description
+
+The transaction id (XID) is present on the request header.
```
galderz: Hmmm, I'm fairly sure in recent versions that was removed since we're not yet using it.

Hot-Rod-Transactions-NonXA.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+|OnePhaseCommit
+|byte
+|When it is set to `1`, the server will use one-phase-commit if available (XA only)
+
+|Number of keys
+|vInt
+|The number of keys
+
+3+^|For each key
```
galderz: I'm not sure I understand why we have to send all of this for each key involved in tx. I mean, a client has executed a put/remove...etc within tx and each of those have a message id. Couldn't you just send those message ids in the prepare function? It would require the server to track those messages but it would make the prepare lighter? Just wondering :)

rvansa: I suppose the tx client wouldn't send the modifications over the wire - it would just fetch values if needed (as originator in tx cache does) and apply the command locally, sending the mods list only once during the prepare.

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: I've noticed and I've put it back :)

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: hmm, it is a good idea.

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: not anymore. I have discussed with Dan (somewhere in this PR) and it is better to mark the transaction as a client transaction to avoid rollback if the originator server node crashes. In this way, I don't need the recovery enabled.

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: ok, I'll change

Hot-Rod-Transactions-NonXA.asciidoc
pruivo: @rvansa is right. All the operations are local to client and it only contacts the server if the value is needed.

galderz:
I guess we can integrate this by now?

pruivo:
yes I think so :)

rvansa:
I still think that we should at least clarify relation of transactions to other features we provide: script execution, querying... (is there something else)? I guess that the plan now is to execute those out of transaction, but I would appreciate if the design was open to changing that while preserving backwards compatibility.
E.g. eventually I'd like to support functional API over Hot Rod, and that will likely utilize scripts. We don't want to close door before executing several scripts in a transaction. If we just ignore the problem, saying that scripts won't run in tx, it means that we can't later come and make them a part of the tx, because that would change semantics.

PR #8, opened 2017-05-22
ConcurrentAPI and InfinispanAPI Object

karesti:
Clustered locks https://issues.jboss.org/browse/ISPN-8069

### DIFF
--- a/ConcurrentAPI.adoc
+++ b/ConcurrentAPI.adoc
@@ -1,0 +1,283 @@
+== Clustered and Concurrent API
+
+=== Motivation
+
+Coming from an initial issue where we need to provide an implementation for vert.x cluster manager for the shared data feature,
+provide concurrent and clustered programming capabilities.
+
+* Clustered Locks
+* Clustered CountDownLatch
+
+=== Scope
+
+All these interfaces should be placed in commons, in order to be used either in Embedded and Server mode.
+
+== Infinispan Clustered Locks
+
+=== ClusteredLockManager and configuration
+
+```java
+package org.infinispan.lock.api;
+
+public class ClusteredLockManager {
+   boolean defineLock(String name, LockConfiguration configuration);
+
+   InfinispanLock get(String name);
+
+   LockConfiguration getConfiguration(String name);
+
+   boolean isDefined(String name);
+
+   CompletableFuture<Boolean> remove(String name);
+
+   CompletableFuture<Void> reset(String name);
+}
+
+
+public class LockConfiguration {
+   private final RentrancyLevel lentrancyLevel; // default NOT_REENTRANT
+
+   private final boolean silentFailover; // default true
+
+   // The maximum length of time for which a client can hold and renew a lock aquisition
+   private final long maxLeaseTime;
+
+   // Maximum length of time a lock may be held without updating the lease,
+   // after that time any attempt to lock it will succeed
+   private final long renewalLeaseTime;
+}
+
+public enum RentrancyLevel {
+   NODE, // Node can lock if it owns the lock without blocking, only the owner node can unlock
+   INSTANCE, // Instance can lock multiple times if it owns the lock without blocking, only the owner instance can unlock
+   NOT_REENTRANT // Nobody can take the lock if already taken, but everybody can release it
+}
+
+@NODE  : when a node owns the lock,
+```
+###### InfinispanLock defineLock(String name, LockConfiguration configuration)
+
+Defines a lock with the specific name and LockConfiguration.
+It does not overwrite existing configurations.
+
+Returns true if successfully defined or false if the lock is already defined or any other failure.
+If silentFailover is false, then InfinispanLockException will be raised.
+
+###### InfinispanLock get(String name)
+
+Get's a InfinipanLock by it's name and throws InfinispanLockException if the lock is not not defined.
+User must call defineLock before this method.
+
+###### Optional<LockConfiguration> getConfiguration(String name);
+
+Get's the configuration for a Lock. If the Lock does not exist, Optional.empty() will be returned.
+
+###### boolean isDefined(String name)
+True if the lock exists, false if it doesn't
+
+###### CompletableFuture<Boolean> remove(String name)
+Removes a Lock from the system. Returns true when it was removed, false when the lock does not exist.
+If any other Runtime problems appear, InfinispanLockException will be raised withe the reason.
+As Locks are not removed automatically, so this has to be done programatically when the Lock is
+no longer needed. Otherwise, OutOfMemoryException could happen.
+
+Remove must be executed when the lock is locked, because running that without exclusive access should
+result in an exception.
+Internally, the implementation should contain generation number so that attempts to acquire a lock
+of a removed generation will result it exceptions in the other callers, too.
+
+###### CompletableFuture<Void> reset(String name)
+Resets the lock to its initial state.
+If any parties are currently waiting at the lock, they will return with failure on the CompletableFuture
+
+=== InfinispanLock
+
+When a cluster node holding a Lock dies, this lock is released and available for the others.
+
+```java
+public interface InfinispanLock {
+
+   CompletableFuture<Void>	lock();
+
+   CompletableFuture<Boolean>	tryLock();
+
+   CompletableFuture<Boolean>	tryLock(long time, TimeUnit unit);
+
+   CompletableFuture<Void>	unlock();
+}
+
+```
+
+###### CompletableFuture<Void> lock();
+
+CompletableFuture is completed successfully when the lock is acquired
+When a lock is aquired by a client, it will be automatically released after the maxLeaseTime specified.
+RenewalLeaseTime is the interval time is the time a client can aquire a lock consecutively
+User should set the timeouts to non-positive value
+The initial embedded implementation does not have to support positive values
+
+###### CompletableFuture<Boolean> tryLock();
+
+Acquires the lock only if it is free at the time of invocation.
+Acquires the lock if it is available and returns with the value true.
+If the lock is not available then this method with the value false.
+
+###### CompletableFuture<Boolean> tryLock(long time, TimeUnit unit);
+
+Acquires the lock if it is free within the given waiting time.
+If the lock is available this method returns with the value true.
+
+Parameters:
+time - the maximum time to wait for the lock
+unit - the time unit of the time argument
+Returns:
+true if the lock was acquired and false if the waiting time elapsed before the lock was acquired
+
+CompletableFuture fails with InfinispanLockException in case of error (InterruptedException, or any other non checked exceptions)
+
+###### CompletableFuture<Boolean> unlock();
+
+If the lock is rentrant (Node or Instance), only the instance or node holding the lock will be able to unlock, otherwise, anybody can unlock
+and it will behave as a Semaphore with one permit.
+True answer will say that the operation was succesul and the lock has been released, false the lock has not been relased
+
+
+=== Demo
+
+This main is an example on calling the API as if it was blocking, the corrent implementation should be done
+
+```java
+   public static void main(String[] args) throws Exception {
+        EmbeddedCacheManager cm = Infinispan.createClustered();
+        CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(cm);
+        counterManager.defineCounter("counter", ...);
+        WeakCounter counter = counterManager.weakCounter("counter");
+        ClusteredLockManager lockManager = EmbeddedLockManagerFactory.asClusteredLockManager(cm);
+        lockManager.defineLock("lock", ...);
+        InfinispanLock lock = lockManager.get("lock");
+
+       for (int i = 0; i < 100; i++) {
+            System.out.println("Counter on " + i + " is => " + counter.getValue());
+            lock.lock()
+               .thenRun(new CounterExample(counter))
+               .whenComplete((nil, t) -> lock.unlock());
+        }
+        cm.stop();
+    }
+
+    static class CounterExample implements Runnable {
+         private WeakCounter counter;
+
+         public CounterExample(WeakCounter counter) {
+            this.counter = counter;
+         }
+
+         @Override
+         public void run() {
+            counter.increment();
+            try {
+               Thread.sleep(1000);
+            } catch (InterruptedException e) {
+               e.printStackTrace();
+            }
+            counter.decrement();
+         }
+      }
+```
+
+
+
+== (preview) Distributed CountDownLatch
+
+A synchronization aid that allows one or more threads to wait until a set of operations being performed in other threads completes.
+
+
+https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html
+
+
+=== CountDownLatchManager and configuration
+
+```java
+
+public class CountDownLatchManager {
+   InfinispanCountDownLatch define(String name, CountDownLatchConfiguration configuration);
+   InfinispanCountDownLatch get(String name);
+   boolean remove(String name);
+}
+
+public interface InfinispanCountDownLatch {
+   CompletableFuture<Void> await();
+   CompletableFuture<Boolean> await(long timeout, TimeUnit unit);
+   CompletableFuture<Long> getCount();
+   void countDown();
+}
+
+
+public class CountDownLatchConfiguration {
+   private long count;
+   private boolean autoRemove;
+}
+
+ * @name A unique name
+ * @count initial count. If the countDownLatch already exists, override the current value
+ * @autoRemove when the countdown is zero, destroy it automatically.
+
+```
+
+=== Demo, just as if was locking interface
+
+```java
+
+   public class Driver {
+
+      public static void main(String[] args) throws InterruptedException {
+
+         EmbeddedCacheManager cm = Infinispan.createClustered();
+         InfinispanCountDownLatch startSignal = cm.countDownLatch("startSignal", CountDownLatchConfig.count(1).create().build());
+         InfinispanCountDownLatch doneSignal = cm.countDownLatch("doneSignal", CountDownLatchConfig.count(1).create().build());
+
+         new Thread(new Worker()).start();
+
+         System.out.println("Do something on driver");
+         Thread.sleep(5000);     // don't let run yet
+         System.out.println("End something on driver");
+         startSignal.countDown();      // let all threads proceed
+         System.out.println("Do something on driver");
+         Thread.sleep(5000);
+         System.out.println("End something on driver");
+         boolean await = doneSignal.await(10, TimeUnit.SECONDS);// wait for all to finish
+         if(await)
+            System.out.println("END DRIVER");
+
+         Infinispan.stopAll();
+      }
+
+      static class Worker implements Runnable {
+         private final InfinispanCountDownLatch startSignal;
+         private final InfinispanCountDownLatch doneSignal;
+         Worker() {
+            EmbeddedCacheManager cm = Infinispan.createClustered();
+            this.startSignal = cm.countDownLatch("startSignal");
+            this.doneSignal = cm.countDownLatch("doneSignal");
+         }
+
+         public void run() {
+            try {
+               boolean await = startSignal.await(4, TimeUnit.SECONDS);
+               if (await)
+                  doWork();
+               else
+                  System.out.println("work could not be done !");
+
+               doneSignal.countDown();
+            } catch (InterruptedException ex) {} // return;
+         }
+
+         void doWork() throws InterruptedException {
+            System.out.println("Start work");
+            Thread.sleep(10000);
+            System.out.println("End work");
+         }
+      }
+   }
+```
--- a/InfinispanAPIObject.adoc
+++ b/InfinispanAPIObject.adoc
@@ -1,0 +1,36 @@
+== Infinispan API
+
+=== Motivation
+
+Being able to offer a simple API to create clustered caches and improve API usability
+
+
+== Infinispan
+
+This factory will help to create a cluster of Infinispan.
+
+Whenever a new member is added on the VM, the implicit configuration will handle the visibility between members.
+
+```java
+
+public final class Infinispan {
+
+    private Infinispan() {}
+
+    //Constructs and starts a new instance of the CacheManager, using the system defaults.
+    public static EmbeddedCacheManager createClusteredCache() {}
+
+    public static EmbeddedCacheManager createClusteredCache(Configuration configuration) {}
+
+    public static EmbeddedCacheManager createClusteredCache(GlobalConfiguration globalConfig, Configuration configuration) {}
+
+    // Return the running CacheManagers on the cluster in the calling VM
+    public static Set<EmbeddedCacheManager> getAllMembers() {}
+
+    // Stop all the members on the VM
+    public static void stopAll() {}
+}
+
+```
+
+


ConcurrentAPI.adoc
```diff
@@ -0,0 +1,220 @@
...
+Coming from an initial issue where we need to provide an implementation for vert.x cluster manager for the shared data feature, provide concurrent and distributed programming capabilities.
+
+* Distributed locks -> is it a good idea ?
+* Distributed semaphores : first use case, vert-x cluster management
+* Distributed Countdown latch : it seems the team has already spoken about having it
+* Distributed Atomic Variables : we already have them, would be worth to offer an specific AtomicLong (for example) ?
+
+=== Scope
+
+All these interfaces should be placed in commons, in order to be used either in Embedded and Server mode.
```
slaskawi: I would also say that they could pair up (or extend) JDK interfaces from `java.util.concurrent`.

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,220 @@
...
+
+== Infinispan Distributed Locks
+
+I'm wondering if we need to expose Locks on a API. Not sure, because we can already Lock a key on the Cache API and I believe this could be confusing.
```
slaskawi: How about providing a single factory for all distributed toys? Something like:

```
public interface ClusteredUtilsFactory {
   Lock distributedLock();
   Semaphore distributedSemaphore();
   .... other toys ....
}
```

InfinispanAPIObject.adoc
```diff
@@ -0,0 +1,36 @@
+== Infinispan API
+
+=== Motivation
+
+Being able to offer a simple API to create clustered caches and improve API usability
```
slaskawi: I think this type of code belongs to the user space. I wouldn't place it in Infinispan (but on a modules that interfaces Infinispan to Vert.x for example).

karesti: Well, since competitors offer this kind of super class and I discovered it because I've playing with their demos, I really think it does belong in our code

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,220 @@
+== Distributed Concurrent API
+
+=== Motivation
+
+Coming from an initial issue where we need to provide an implementation for vert.x cluster manager for the shared data feature, provide concurrent and distributed programming capabilities.
+
+* Distributed locks -> is it a good idea ?
+* Distributed semaphores : first use case, vert-x cluster management
+* Distributed Countdown latch : it seems the team has already spoken about having it
+* Distributed Atomic Variables : we already have them, would be worth to offer an specific AtomicLong (for example) ?
```
tsegismont: Can you point me to this? How do they differ from JGroups counters?

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,220 @@
...
+== Infinipan Distributed Semaphores
+
+Distributed version of java.util.concurrent.Semaphore.
+
+http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html
+
+To create a Semaphore, use the Infinispan Utility class, passing the semaphoreConfig or using the default values or XML values.
+Configuration can be ovrerriden programatically.
+
+First Semaphore wont't implement fairness, this can be cosidered after.
```
tsegismont: FYI this is not required by Vert.x

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,220 @@
...
+
+   int availablePermits();
+
+   int drainPermits();
+
+   void reducePermits(int reduction)
+
+   void release();
+
+   void release(int permits);
```
tsegismont: For Vert.x it's important that those methods can be called by any thread, not just the thread which acquired the permits.

karesti:
@tsegismont yes ! We are going to rethink most of it and simplify, I'm working on it and won't take long. I believed that the permits on the semaphore were indeed changed sometimes, so I was confused about it, but I understood better now 💃

ConcurrentAPI.adoc
karesti: Sure 👍 thank you for the clarification ! I had it wrong

ConcurrentAPI.adoc
karesti: Ok. Anyway, we are not going to start with a semaphore yet, but a distributed async lock that could be released by any thread. So like a semaphore with 1 permit.

ConcurrentAPI.adoc
karesti: We have some AtomicObjectFactory, I don't know the details yet and apparently is a bit experimental

karesti: http://blog.infinispan.org/2015/01/a-factory-of-atomic-objects.html

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
+== Distributed Concurrent API
+
+=== Motivation
+
+Coming from an initial issue where we need to provide an implementation for vert.x cluster manager for the shared data feature,
+provide concurrent and distributed programming capabilities.
+
+* Distributed Locks
```
tristantarrant: We had a discussion a while ago about what should be called distributed and what should we call clustered.
For future reference:
* distributed: information which is divided/sharded among the nodes in the cluster.
* clustered: information which is global to the cluster (owned by a single node or by all nodes)

karesti:
@tsegismont there is a new API called CounterManager implemented by @pruivo mostly, that already does this atomic long style (I didn't know about this API before). This API is available since Infinispan 9.0 so the AtomicLongs are not necessary. I think we can use it in vert-x infinispan and see how it works

InfinispanAPIObject.adoc
tristantarrant: I agree with Katia: our API is not as usable / discoverable as it could be.

tristantarrant:
@karesti @tsegismont CounterManager will be available from 9.1.0.Alpha1

karesti:
@tristantarrant I thought it was already available looking at the code ! 👯

karesti:
@tristantarrant can we still give a try as a preview ?

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+Defines a lock with the specific name and LockConfiguration.
+It does not overwrite existing configurations.
+
+Returns true if successfully defined or false if the counter exists or any other failure.
+If silentFailover is false, then InfinispanLockException will be raised.
+
+###### InfinispanLock get(String name)
+
+Get's a InfinipanLock by it's name and creates it if does not exist with the default configuration.
```
galderz: I'm not sure about this create-if-not-exist. This is how `Cache.getCache()` worked and we've come to regret this create-if-not-exist behaviour. I'd rather separate into create/get methods. Maybe defineLock can simply be called `createLock`?

InfinispanAPIObject.adoc
slaskawi: Ok understood.

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+public interface InfinispanCountDownLatch {
+   CompletableFuture<Void> await();
+   CompletableFuture<Boolean> await(long timeout, TimeUnit unit);
+   CompletableFuture<Void> countDown();
+   CompletableFuture<Long> getCount();
+}
+
+
+public class CountDownLatchConfiguration {
```
rvansa: Since this does not appear anywhere else, you could well inline it into `define` parameters

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   InfinispanCountDownLatch get(String name);
+   boolean remove(String name);
+
+}
+
+public interface InfinispanCountDownLatch {
+   CompletableFuture<Void> await();
+   CompletableFuture<Boolean> await(long timeout, TimeUnit unit);
+   CompletableFuture<Void> countDown();
+   CompletableFuture<Long> getCount();
```
rvansa: count in config is integer

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+public class CountDownLatchConfiguration {
+   private String name;
+   private int count;
+   private boolean createIfAbsent;
+   private boolean autoRemove;
+}
+
+ * @name A unique name
+ * @count initial count. If the countDownLatch already exists, override the current value
```
rvansa: Uuuh, wouldn't it be safer to throw an exception?

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+
+=== CountDownLatchManager and configuration
+
+```java
+
+public class CountDownLatchManager {
+
+   InfinispanCountDownLatch define(CountDownLatchConfiguration configuration);
+   InfinispanCountDownLatch get(String name);
```
rvansa: I hope this won't create-if-absent

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+}
+
+ * @name A unique name
+ * @count initial count. If the countDownLatch already exists, override the current value
+ * @createIfAbsent if it does not exist, and the flag is set to true, then create it with the name and count
+ * @autoRemove when the countdown is zero, destroy it automatically.
+
+```
+
+=== Demo, just as if was locking interface
```
rvansa: Since you'd like to focus on clustered case, you could split this into two classed (Driver/Worker) with `main` method.

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+         new Thread(new Worker()).start();
+
+         System.out.println("Do something on driver");
+         Thread.sleep(5000);     // don't let run yet
+         System.out.println("End something on driver");
+         startSignal.countDown();      // let all threads proceed
+         System.out.println("Do something on driver");
+         Thread.sleep(5000);
+         System.out.println("End something on driver");
+         boolean await = doneSignal.await(10, TimeUnit.SECONDS);// wait for all to finish
```
rvansa: demo outdated, interface uses CF

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+public class CountDownLatchManager {
+
+   InfinispanCountDownLatch define(CountDownLatchConfiguration configuration);
+   InfinispanCountDownLatch get(String name);
+   boolean remove(String name);
+
+}
+
+public interface InfinispanCountDownLatch {
+   CompletableFuture<Void> await();
```
rvansa: Either the use of `CompletableFuture` or `await` does not sound right, as this latter evokes blocking and it could be easily misused. Maybe 'asyncWait`?
The second overload won't be needed with Java 9's `orTimeout` but it's useful for now. Just make sure that you don't create `ScheduledThreadPool` per instance :)

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   CompletableFuture<Boolean> await(long timeout, TimeUnit unit);
+   CompletableFuture<Void> countDown();
```
rvansa: Not sure about the need to return a CF here - I think that this should be a fire & forget one. In case of an exception, the implementation should propagate the failure to all waiting latches, rather than forcing the user to deal with that (how should he?).

ConcurrentAPI.adoc
rvansa: I was looking into these atomic objects implementation and woooh... It uses cache.put() calls and event listeners basically as a multicast messaging API, with guaranteed delivery order.

It does not keep the actual values in the cache, user code is expected to hold(or lose) all replicas. So contrary to the blogpost, I don't think that this is
> making transparently the object replicated and durable

(speaking about durability: this does not faciliate cache stores)

While I could ignore that listeners might not be 100% reliable and the code does not handle retries, the 'responses' are always multicast as well. It's simply wrong tool for the job.

ConcurrentAPI.adoc
karesti: Then Clustered should be better here

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+public class DistributedLockManager {
+   boolean defineLock(LockConfiguration configuration);
+   InfinispanLock get(String name);
+   LockConfiguration getConfiguration(String name);
+   boolean isDefined(String name);
+   CompletableFuture<Boolean> remove(String name);
+}
+
+
+public class LockConfiguration {
```
karesti: Not as an start point. We can consider to implement fair locks after

ConcurrentAPI.adoc
karesti: @galderz @rvansa I called it "define" to have coherent API. The new Counters API does work like that (org.infinispan.counter.api.CounterManager). I guess this method names were stated last year. But I do agree on the "create-if-not-exists". Counter API works indeed in this way First user defines a counter and after get's the counter.

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+== Infinispan Distributed Locks
+
+=== DistributedLockManager and configuration
+
+```java
+package org.infinispan.lock.api;
+
+public class DistributedLockManager {
+   boolean defineLock(LockConfiguration configuration);
```
karesti: Absolutely

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   InfinispanLock get(String name);
```
karesti: well, for me this should behave like counters somethow. Why not making everything non blocking. I don't think this is a big issue concerning vert.x

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   LockConfiguration getConfiguration(String name);
+   boolean isDefined(String name);
+   CompletableFuture<Boolean> remove(String name);
+}
```
karesti: I agree. So you mean adding another method to do that on the API ? Or checking somehow management rights to do it ?

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+
+public class LockConfiguration {
+   private final String name; // default "ISPN-Lock"
+   private final boolean rentrant; // default true
```
karesti: @rvansa vert.x needs a non rentrant lock for their SharedData functionality. Internally, vert-x hazelcast cluster manager uses a semaphore with a single permit to achieve this. At first I was about to propose that we implement distributed semaphores. But after a discussion with Will, we decided to offer a Lock. This lock can handle different levels of rentrancy. So for vert-x usage, would behave as a semaphore with a single permit, but we can offer our users clustered locks rentrant per thread, or rentrant per cluster node for example. And we can start with a non rentrant implementation that will be used by vert-x/infinispan cluster manager.

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   private final boolean silentFailover; // default true
+}
+
+```
+###### InfinispanLock defineLock(LockConfiguration configuration)
+
+Defines a lock with the specific name and LockConfiguration.
+It does not overwrite existing configurations.
+
+Returns true if successfully defined or false if the counter exists or any other failure.
```
karesti: define means create ... it's confusing, I agree

ConcurrentAPI.adoc
rvansa: It seems to me that it is a big issue. Either you call `vertx.executeBlocking`(which you should avoid) or execute a blocking call in a thread that should not get blocked. Or you make sure that this `get(...)` method never blocks, by not doing any RPCs. Please choose.

ConcurrentAPI.adoc
rvansa: At this point add a new method into the interface. Exposure via JMX could be defined later on.

ConcurrentAPI.adoc
karesti: I agree this is a valid concern, will pointed to this issue when we discussed last week.
However, we must choose. Our API's should be homogeneous and counters don't work this way right now. @pruivo not sure you follow everything we are saying here, but as the CounterManager API is not official yet, we might be discussing together the public API design so these API are similar for the infinispan users.

ConcurrentAPI.adoc
rvansa: Briefly looking into the impl it seems that CounterManager.defineCounter returns false if counter **configuration** exists, not the actual counter. @pruivo ?
Though, I am having second thoughts about the create-if-not-exist; locks are lightweight objects and the synchronization through cache isn't that complex.

karesti:
@rvansa I did the examples as blocking and I had no time to make them and review some things before sharing, I wanted @wburns to check before I left. 
I'm going to take all your reviews and correct some of the misunderstandings, typo bugs and other stuff that are not very clear right now, including demos.

ConcurrentAPI.adoc
karesti: yes, I think so. I just adapted a simple example to illustrate, but for the infinispan simple examples I think 2 main classes could be better :)

ConcurrentAPI.adoc
karesti: ok

ConcurrentAPI.adoc
karesti: ok

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+=== Demo (as if the code was blocking)
+
+This main is an example on calling the API as if it was blocking, the corrent implementation should be done
+
+```java
+   public static void main(String[] args) throws Exception {
+        EmbeddedCacheManager cm = Infinispan.createClustered();
+        CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(cm);
+        DistributedLockManager lockManager = EmbeddedLockManagerFactory.asDistributedLockManager(cm);
+        InfinispanLock lock = lockManager.createLock("lock");
```
karesti: define :D

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+CompletableFuture fails with InfinispanLockException in case of error (InterruptedException, or any other non checked exceptions)
+
+###### CompletableFuture<Boolean> unlock();
+
+If the lock is rentrat, only the Thread holding the lock will be able to unlock, otherwise, any thread can unlock
+and it will behave as a Semaphore with one permit.
+true answer will say that the operation was succesul and the lock has been released, false the lock has not been relased
+
+
+=== Demo (as if the code was blocking)
```
karesti: because it was already done and anyway, this is not the final demo in any case

ConcurrentAPI.adoc
karesti: something like CompletableFuture<Boolean> unlock(String name); ?

karesti: I added unlock

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+Get's a InfinipanLock by it's name and creates it if does not exist with the default configuration.
+If you need to change the default configuration - creating a non rentrant lock for example, then call
+defineLock first.
+
+###### Optional<LockConfiguration> getConfiguration(String name);
+
+Get's the configuration for a Lock. If the Lock does not exist, Optional.empty() will be returned.
+
+###### boolean isDefined(String name)
+True if the lock exists, false if it doesn't
```
karesti: defined

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+###### CompletableFuture<Boolean> remove(String name)
```
karesti: I'm taking this comment on the doc

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+Removes a Lock from the system. True when it was removed, false when the lock does not exist.
+If any other Runtime problems appear, InfinispanLockException will be raised withe the reason.
+As Locks are not removed automatically, so this has to be done programatically when the Lock is
+no longer needed. Otherwise, OutOfMemotyException could happen.
+
+=== InfinispanLock
```
karesti: I put the maxLeaseTime and renewalLeaseTime in configuration

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   CompletableFuture<Boolean>	tryLock(long time, TimeUnit unit);
+
+   CompletableFuture<Void>	unlock();
+}
+
+```
+
+###### CompletableFuture<Void> lock();
+
+Acquires the lock.
```
karesti: ok

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,255 @@
...
+
+   CompletableFuture<Void>	unlock();
+}
+
+```
+
+###### CompletableFuture<Void> lock();
+
+CompletableFuture is completed successfully when the lock is acquired.
+In client/server mode, when a lock is aquired by a client, it will be automatically released
```
rvansa: Uh, so the lease times apply only for client/server and not embedded? These shouldn't differ, if embedded should be everlasting (until the node dies), user should set the timeouts to non-positive value (to be consistent with our expiration settings). The initial embedded implementation does not have to support positive values.

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,255 @@
...
+time - the maximum time to wait for the lock
+unit - the time unit of the time argument
+Returns:
+true if the lock was acquired and false if the waiting time elapsed before the lock was acquired
+
+CompletableFuture fails with InfinispanLockException in case of error (InterruptedException, or any other non checked exceptions)
+
+###### CompletableFuture<Boolean> unlock();
+
+If the lock is rentrant, only the Thread holding the lock will be able to unlock, otherwise, any thread can unlock
```
rvansa: no threads, please

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,255 @@
...
+
+public class LockConfiguration {
+   private final RentrancyLevel lentrancyLevel; // default None
+   private final boolean silentFailover; // default true
+   private final long maxLeaseTime; // after a time the lock is released in milles
+   private final long renewalLeaseTime; // time after a client can aquire the lock
+}
+
+public enum RentrancyLevel {
+   None, Node, Thread
```
rvansa: I would also suggest `NOT_REENTRANT`, since `NONE` and `NODE` are easy to misspell/misread.

ConcurrentAPI.adoc
rvansa: yes

ConcurrentAPI.adoc
rvansa: In other words, I would make the defaults `-1` and support only those in initial embedded implementation, but later the remote client should not support non-positive (because that would lead to deadlocks).

karesti: ok

ConcurrentAPI.adoc
karesti: ok

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,255 @@
...
+Internally, the implementation should contain generation number so that attempts to acquire a lock
+of a removed generation will result it exceptions in the other callers, too.
+
+###### CompletableFuture<Void> unlock(String name)
+Unlocks a lock in the system. This might be useful to unlock a lock from a management point of view if
+needed to.
+
+=== InfinispanLock
+
+When a instance holding a Lock dies, this lock is released and available for the other instances.
```
karesti: ok

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,255 @@
...
+###### boolean isDefined(String name)
+True if the lock exists, false if it doesn't
+
+###### CompletableFuture<Boolean> remove(String name)
+Removes a Lock from the system. Returns true when it was removed, false when the lock does not exist.
+If any other Runtime problems appear, InfinispanLockException will be raised withe the reason.
+As Locks are not removed automatically, so this has to be done programatically when the Lock is
+no longer needed. Otherwise, OutOfMemoryException could happen.
+
+Remove should be executed when the lock is locked, because running that without exclusive access should
```
karesti: yep

ConcurrentAPI.adoc
karesti: Still talking about threads because I'm very persistent in my mistakes 🗡 
correcting it

karesti: How many rentrancy levels do you imagine here then ?

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,255 @@
...
+   CompletableFuture<Boolean> remove(String name);
+   CompletableFuture<Void>	unlock(String name);
+}
+
+
+public class LockConfiguration {
+   private final RentrancyLevel lentrancyLevel; // default None
+   private final boolean silentFailover; // default true
+   private final long maxLeaseTime; // after a time the lock is released in milles
+   private final long renewalLeaseTime; // time after a client can aquire the lock
```
karesti: Ok, I understood this names the other way around. Updating comments

ConcurrentAPI.adoc
karesti: corrected

ConcurrentAPI.adoc
```diff
@@ -0,0 +1,244 @@
...
+   CompletableFuture<Boolean>	tryLock();
+
+   CompletableFuture<Boolean>	tryLock(long time, TimeUnit unit);
+
+   CompletableFuture<Void>	unlock();
+}
+
+```
+
+###### CompletableFuture<Void> lock();
```
karesti: it should wait for me ...

karesti: and maybe throw and exception after a specific amount of time ?

ConcurrentAPI.adoc
rvansa: Sorry, I wasn't clear: `NOT_REENTRANT`, `NODE` and `INSTANCE`. Just didn't want two modes to differ by single letter only.

karesti: ok, cool

ConcurrentAPI.adoc
rvansa: For timeouts, you have `tryLock()`. I assume 'wait for me' means deadlock.

ConcurrentAPI.adoc
karesti: hm... ok

ConcurrentAPI.adoc
galderz: Hmmmm, I guess a lock is a not a cache with it's start/stop lifecycle methods... I'm happy to keep names consistent with CounterManager.

InfinispanAPIObject.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+Whenever a new member is added on the VM, the implicit configuration will handle the visibility between members.
+
+```java
+
+public final class Infinispan {
+
+    private Infinispan() {}
+
+    //Constructs and starts a new instance of the CacheManager, using the system defaults.
```
tristantarrant: the latter

InfinispanAPIObject.adoc
```diff
@@ -0,0 +1,36 @@
...
+    public static EmbeddedCacheManager createClusteredCache() {}
```
rvansa: Assuming that the InfinispanAPI will be an access point for both caches, locks etc., I wonder if we want a static method `createCacheManager()`. I am wondering if we don't rather want Infinispan object bound to one cache manager, and relaying the other builiding blocks to the same CM:
```java
class Infinispan {
   static Infinispan newInstance();
   EmbeddedCacheManager caches();
   ClusteredLockManager locks();
   ...
}
```

tristantarrant: This is the style of interface I'd like to see.
Additionally I'd like to see a way to piggy-back a  new instance of "Infinispan" on an existing one, so things like transport, thread-pools, etc can be shared:
```java
static Infinispan newInstance(Infinispan other);
```

rvansa: @tristantarrant Why would you want to have a duplicate instances? I see `Infinispan` as just a wrapper over CM, the instance should not have any state on its own.

tristantarrant: Indeed, my intention was never to imply that `Infinispan` would have its own state. And while it is already possible for multiple instance for DCM to share pieces, I was thinking of making that process a bit cleaner.

PR #9, opened 2017-07-03
Transcoding details

gustavonalle:
From encoders to indexable/queryable Json

### DIFF
--- a/Compatibility-2.0.md
+++ b/Compatibility-2.0.md
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
 We must ensure that Consistent-Hash-aware clients knowledge of the hashing of keys is not affected by transcoding.
 
 Transcoders should be deployable
+
+
+# Encoders, Transcoders and indexable Json 
+
+[ISPN-7896](https://issues.jboss.org/browse/ISPN-7896) introduced the Encoders, adding capability to the caches for on-demand data conversion instead of relying on implicit conversions based on a combination of flags and interceptors. 
+
+This section will expand on the Encoders usage and will introduce the complementary Transcoders, used to convert between different encodings.
+
+## About Encoders
+
+An Encoder is essentially a two-way conversion between the storage format in a cache and the read/write format. The ```AdvancedCache```  interface has a method to specify an encoder to be used during all operations involving that cache:
+
+```cache.getAdvancedCache().withEncoding(MyEncoder.class)```
+
+which returns a decorated cache instance where all operations (read/write/stream/add listener, etc) flow through the supplied encoder.
+
+At the moment, there is no cache configuration to specify an Encoder; Infinispan tries to pick the best decorated cache depending on the
+situation where this cache is used. For example:
+
+* Embedded caches with ```Storage.OBJECT``` are associated with the ```IdentityEncoder```, which stores objects "as is";
+* Embedded caches with ```Storage.OFF_HEAP``` are associated with a ```MarshallerEncoder```, which marshalls data before storing in order to produce a ```byte[]``` (hard requirement);
+* Embedded caches with ```Storage.BINARY``` are associated with the BinaryEncoder, which produces a mix of marshalled and unmarshalled objects.
+* Caches used in the server with indexing enabled and compatibility mode enabled uses a ```ProtostreamCompatEncoder```, where the byte[] arriving from the client
+ is unmarshalled via a protostream marshaller into a java representation.
+* Caches used in the server with no indexing and compatibility use the ```IdentityEncoder```, which treats content read and written as opaque ```byte[]```.
+* Caches used in the server with compatibility (and no indexing) uses the CompatModeEncoder, which takes the ```byte[]``` from the client and unmarshalls them to java object for storage.
+* Scripts which carry data-type ```utf-8``` wraps a cache with the ```UTF8Encoder```, that will extract ```byte[]``` out of a String in order to store and will create Strings out of ```bytes[]``` for reading
+operations.
+
+
+Some endpoints in the Infinispan Server allow to send and request data in a certain format (MimeType):
+
+* Scripting: scripts (.js at least) can pass a header "dataType" with a MimeType string;
+* Rest: request headers "content-type" (for PUT/POST) and "accept" (for GET) can carry mime types as well
+* HotRod: there is no such capability yet, but it's something that will eventually be added.
+
+
+How does the Mimetype relate to the encoder general mechanism?
+
+
+## Proposal
+
+The proposal is to use a pair (Encoder, MimeType) to describe cache storage. Examples: 
+
+* "```application/json``` with no encoding" means IdentityEncoder associated of a content that represents a json (```byte[]```, ```Gson```, ```Map<String, Map<?,?>>```, etc)
+* "```text/plain; charset=utf-8``` with marshalling" means content produced by the MarshallerEncoder that represents text, so it's a ```byte[]```
+* "```application/octet-stream``` with KryoEncoder" means ```byte[]``` produced by a KryoMarshaller.
+
+This means that different MimeTypes can be associated with the same Encoder, e.g. the IdentityEncoder can be associated with any MimeType meaning whatever is sent
+to cache will be written or read without transformations.
+
+The purpose of the MimeType is twofold:
+
+* Provide a metadata about the content of the cache, independently of the encoding;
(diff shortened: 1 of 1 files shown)


Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+
+# Encoders, Transcoders and indexable Json 
+
+[ISPN-7896](https://issues.jboss.org/browse/ISPN-7896) introduced the Encoders, adding capability to the caches for on-demand data conversion instead of relying on implicit conversions based on a combination of flags and interceptors. 
+
+This section will expand on the Encoders usage and will introduce the complementary Transcoders, used to convert between different encodings.
+
+## About Encoders
+
+An Encoder is essentially a two-way conversion between the storage format in a cache and the read/write format. The ```AdvancedCache```  interface has a method to specify an encoder to be used during all operations involving that cache:
```
slaskawi: ... the read/write format (which is transferred through the wire).

Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+How does the Mimetype relate to the encoder general mechanism?
+
+
+## Proposal
+
+The proposal is to use a pair (Encoder, MimeType) to describe cache storage. Examples: 
+
+* "```application/json``` with no encoding" means IdentityEncoder associated of a content that represents a json (```byte[]```, ```Gson```, ```Map<String, Map<?,?>>```, etc)
+* "```text/plain; charset=utf-8``` with marshalling" means content produced by the MarshallerEncoder that represents text, so it's a ```byte[]```
+* "```application/octet-stream``` with KryoEncoder" means ```byte[]``` produced by a KryoMarshaller.
```
slaskawi: I found it just recently in REST, sometimes users don't care about the format. They just use `Content-type: */*`. I think we should consider doing a fallback `text/plain` for example. 

Also you may be interested in making charset optional or even "replaceable". E.g. one might store entry in UTF8 but get it using ASCII. The good news is that the charset problem touches only "text" mime-types. For vast majority things like JSON, `UTF-8` is the only valid encoding.

The point I'm trying to make is that it *may* be n to 1 mapping (n mime-types to 1 converter). Just make sure we support this well enough.

Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+The MimeType will be either present in the config or inferred; if not provided, Infinispan will automatically pick the best 
+decorated cache depending on the situation.
+
+
+### API changes
+Apart from being able to use a different encoders on the fly, it will be possible to ask for a different MimeType using the method:
+
+```java
+// Both keys and types as "application/xml"
+cache.getAdvancedCache.withMimeType("application/xml").withEncoding(IdentityEncoder.class)
```
slaskawi: What about charset here? Of course this only kicks in when we try to use mime type "text/plain" or "application/text".

Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+#### 3. Compressing server
+
+It'd be also possible to transparently compress and de-compress data by using the right configuration:
+
+```xml
+<local-cache>
+  <data-type>
+     <!-- keys are Strings, values are text that will be stored compressed -->
+     <key mime-type="application/x-java-object"/>
+     <value mime-type="text/plain" encoder="org.infinispan.commons.dataconversion.GzipEncoder"/>
```
slaskawi: That's :1st_place_medal: functionality!

Compatibility-2.0.md
gustavonalle: If no content-type is provided, IMO it should return content in the cache's configured mime type (see below for examples).
WRT to charset, the mime type string can take parameters, e.g. ```text/html; charset=UTF-8``` or ```application/x-java-object; type=java.util.Map```, this would allow for variations when asking for an specific content.

Compatibility-2.0.md
gustavonalle: see previous comment

Compatibility-2.0.md
rvansa: The wire transfers bytes according to the Externalizer, the rw format is the format used on the API border.

Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+#### Example
+
+Consider a cache defined as:
+
+```xml
+<!-- value is defined the same way... -->
+<key mime-type="application/x-java-object" encoder="org.infinispan.commons.dataconversion.IdentityEncoder"/>
+```
+
+A ```cache.getAdvancedCache.withMimeType("application/json").put("{1}","{a}")``` will internally do the following transformations:
```
rvansa: This assumes that both key & value share the same MIME type. Do you have/plan overloads?

Compatibility-2.0.md
rvansa: I am slightly confused about the mime-type with encoder: is the mime-type expected to be input format for the encoder, or its output format? Because GzipEncoder would transform text/plain -> application/octet-stream or [application/gzip](https://www.iana.org/assignments/media-types/application/gzip)

gustavonalle: The ```mime-type``` is the input to the encoder and the output from it (when decoding).

The MimeType is the definition of the user content regardless of the encoding mechanism. The encoder may result in a transformation such as the storage format is better described by a different mime type.


The configuration pair ```mime-type``` / ```encoder``` is loosely based in the pair ```Content-Type``` / ```Content-Encoding``` from the HTTP spec. Here's the snippet from the definition of ```Content-Encoding```:

```
The Content-Encoding [...] is used as a modifier to the media-type. When present, its value 
indicates what additional content codings have been applied [...], and thus what decoding 
mechanisms must be applied in order to obtain the media-type referenced by the Content-Type 
header field. Content-Encoding is primarily used to allow a document to be compressed without 
losing the identity of its underlying media type. 
```

Compatibility-2.0.md
gustavonalle: Correct, R/W format is the format from the API user POV, the wire format is whatever ```Encoder#toStorage()``` produces, fed into the cache's marshaller

Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+
+public interface Transcoder {
+
+   /**
+    * Transcode content between two different {@link MimeType}
+    *
+    * @param content         Content to transcode
+    * @param contentType     The {@link MimeType} of the content
+    * @param destinationType The target {@link MimeType} to convert to
+    * @param binary          if true, produces byte[] as output
```
galderz: Couldn't you use the configured encoder to define what the returned type is? Wouldn't that give more flexibility that just a boolean that defines whether a byte[] is produced?

Compatibility-2.0.md
galderz: Mime/Encoding is a great and flexible way to deal with this issue, but let's make sure this is clearly explained when it comes to docu. The content-type/content-encoding is a great comparison.

Compatibility-2.0.md
```diff
@@ -26,3 +26,304 @@ Embedded Mode also might use a decorated Cache using a Transcoder. For example t
...
+```cache.getAdvancedCache().withEncoding(MyEncoder.class)```
+
+which returns a decorated cache instance where all operations (read/write/stream/add listener, etc) flow through the supplied encoder.
+
+At the moment, there is no cache configuration to specify an Encoder; Infinispan tries to pick the best decorated cache depending on the
+situation where this cache is used. For example:
+
+* Embedded caches with ```Storage.OBJECT``` are associated with the ```IdentityEncoder```, which stores objects "as is";
+* Embedded caches with ```Storage.OFF_HEAP``` are associated with a ```MarshallerEncoder```, which marshalls data before storing in order to produce a ```byte[]``` (hard requirement);
+* Embedded caches with ```Storage.BINARY``` are associated with the BinaryEncoder, which produces a mix of marshalled and unmarshalled objects.
```
gustavonalle: This is describing how ```storeAsBinary``` works: it marshalls all pojos but store primitives and Strings as is

PR #10, opened 2018-10-02
ServerNG

tristantarrant:


ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+* `/server/configuration` configuration files
+   * `infinispan.xml`
+   * keystores
+   * `logging.properties` for configuring logging
+   * User/groups property files (e.g. `mgmt-users.properties`, `mgmt-groups.properties`) 
+* `/server/data` data files organized by container name
+   * `default`
+      * `caches.xml` runtime cache configuration
+      * `___global.state` global state
+      * `mycache` cachestore data
```
rvansa: What if a cache has multiple stores? Will we enforce all stores here only? (SIFS would love to have the index on SSD/NVM device)

ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+
+* `security` configures the available security realms which can be used by the endpoints.
+* `cache-container` multiple containers may be configured, distinguished by name.
+* `endpoints` lists the enabled endpoint connectors (hotrod, rest, ...).
+* `socket-bindings` lists the socket bindings.
+
+An example skeleton configuration file looks as follows:
+
+```
+<infinispan>
```
rvansa: Is this supposed to be the same schema as for embedded (will we just ignore the extras when used in embedded mode?), or something extending it?

ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+
+formatter.PATTERN=org.jboss.logmanager.formatters.PatternFormatter
+formatter.PATTERN.properties=pattern
+formatter.PATTERN.pattern=%d{HH:mm:ss,SSS} %-5p [%c] (%t) %s%E%n
+```
+# Internals
+
+The following is a dump of various internal facts about the server, in no particular order:
+
+* All containers handled by the same server share the same thread pools and transport.
```
rvansa: but you show some `<transport ...>` element in the configuration above...

ServerNG.md
tristantarrant: No this is the default location which is used when just using <file-store/>. It can be overridden per-cache in the usual way.

ServerNG.md
tristantarrant: It is the same schema, which is already extensible through additional parsers (see the counters).
One main difference is allowing multiple cache-containers (identified by unique names).

ServerNG.md
tristantarrant: Right, that's a legacy leftover from the current server which allows distinct channels per cache container, but in the name of simplification, that should go.

ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Layout
+
+The server is laid out as follows:
+
+* `/bin` scripts
+   * `server.sh` server startup shell script for Unix/Linux
+   * `server.ps1` server startup script for Windows Powershell 
+* `/lib` server jars
+   * `infinispan-server.jar` uber-jar of all dependencies required to run the server.
```
wfink: Is this only one jar with all the needed classes, or just a jar to reference all the dependencies?
In any case this would make any patching/updates worse as the users want to know what the difference is.
In that case there is no simple way to replace a module or 3rd party like JGroups

ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+* `/server` default server instance folder
+* `/server/log` log files
+* `/server/configuration` configuration files
+   * `infinispan.xml`
```
wfink: Is there only one configuration or do we have different examples and the default is infinispan.xml which contains a simple clustering?

ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+      <socket-binding name="cluster" multicast-address="${infinispan.bind.address.multicast}" multicast-port="45700"/>
+   </socket-bindings>
+</infinispan>
+```
+
+# Logging
+
+Logging is handled by JBoss Logging's LogManager. This is configured through a `logging.properties` file in the 
+`server/configuration` directory. The following is an example:
+
```
wfink: use a properties file is unlikely, it would be better to have a xml configuration for this as it is for wildfly.
Maybe it can be a choice which prefer xml?

ServerNG.md
rvansa: So what happens when you provide a config with these extra elements do embedded mode? Error/warning/ignored?

ServerNG.md
```diff
@@ -0,0 +1,155 @@
...
+* small codebase with as little duplication of already existing functionality (e.g. configuration)
+* embeddable: the server should allow for easy testability in both single and clustered configurations
+* RESTful admin capabilities
+* Logging using [JBoss Logging logmanager](https://github.com/jboss-logging/jboss-logmanager)
+* Security using [Elytron](https://github.com/wildfly-security/wildfly-elytron)
+
+# Layout
+
+The server is laid out as follows:
```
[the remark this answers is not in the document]
wfink: The xsd's should be available at lease as addon for reference

ServerNG.md
tristantarrant: Right, a single jar doesn't provide any advantage.
Patching, however, in the world of docker images, becomes less of an issue, as the user will just pull a more recent image and won't see what is inside it anyway.

ServerNG.md
tristantarrant: Yes, examples go in docs/examples

ServerNG.md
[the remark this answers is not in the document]
tristantarrant: I have not added the docs dir structure yet. Will do soon.

ServerNG.md
tristantarrant: I wanted to minimize the effort of parsing yet another custom structure and just delegating it to something external. Also, logging != configuration in my opinion.

galderz:
@tristantarrant Are you doing more updates to this?

PR #14, opened 2019-03-05
[draft] True Cache API

galderz:
Design draft for a true Cache API.

It's centered around our Hibernate experience, which is the most popular cache user we have for Infinispan. The suggested APIs would not be compatible with current Hibernate cache region APIs but that's fine.

Methods in which values are fetched/updated/removed in the source of truth data source are crucial to how the cache behaves externally, that's why they're integral to the cache API, e.g. `fetcher` function.

I'm not 100% sure about the exact asynchronous aspects. I'll be making some comments in the PR with doubts I currently have.

### DIFF
--- a/TrueCacheAPI.md
+++ b/TrueCacheAPI.md
@@ -1,0 +1,236 @@
+# True Cache API
+
+This design document defines a true Cache API for use cases such as Hibernate 2L cache.
+Throughout the document we will use Hibernate 2L as use case for explaining rationale behind the API.
+
+
+## Read-Only Cache
+The most basic of cache APIs, supporting only read-only data, should allow:
+
+1. Check if data is present in cache, if present return it, otherwise fetch it from source and store it into the cache.  
+This complex operation is the essence of a caching API.
+In Hibernate, this is split into two Hibernate callbacks into the cache region api: 
+`CachedDomainDataAccess.get` and (maybe) `CachedDomainDataAccess.putFromLoad`.
+
+2. Invalidate individual or all cached entries.
+The user might want to invalidate individual or all cache entries in the cache. 
+For example, if a fresher data has been stored into the database via methods other than Hibernate. 
+
+4. Provide an estimate of the number of entries in cache.
+A user might want to find out the estimated number of entries in a cache.
+This is useful for making assertions about the cached contents.
+
+Here's an an asynchronous API that would match these requirements:
+
+```java
+interface ReadOnlyCache<K, V> {
+  
+  /**
+   * Checks if the cache contains a value associated with the key.
+   * If the value is present, it's returned immediately wrapped in a CompletionStage.
+   * 
+   * If the value is not present, the function is called asynchronously to fetch the value from an unknown source.
+   * If the supplier returned a value, the value is cached and then returned to the caller wrapped in a CompletionStage.
+   * 
+   * If the supplier didn't find a value, null is returned to the caller wrapped in a CompletionStage.
+   *  
+   * @param key to look up in the cache
+   * @param fetcher called if not value is associated with the key in the cache
+   * @return a future containing the cached entry if present, otherwise the value returned by the supplier 
+   */
+  CompletionStage<V> get(K key, Function<K, V> fetcher);
+
+  /**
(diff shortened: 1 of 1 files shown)


TrueCacheAPI.md
```diff
@@ -0,0 +1,236 @@
...
+   * If the value is not present, the function is called asynchronously to fetch the value from an unknown source.
+   * If the supplier returned a value, the value is cached and then returned to the caller wrapped in a CompletionStage.
+   * 
+   * If the supplier didn't find a value, null is returned to the caller wrapped in a CompletionStage.
+   *  
+   * @param key to look up in the cache
+   * @param fetcher called if not value is associated with the key in the cache
+   * @return a future containing the cached entry if present, otherwise the value returned by the supplier 
+   */
+  CompletionStage<V> get(K key, Function<K, V> fetcher);
```
galderz: From an asynchronous perspective, should `Function<K, V>` be `Function<K, CompletionStage<V>` instead?

TrueCacheAPI.md
```diff
@@ -0,0 +1,236 @@
...
+   * If the comparator returns a positive number,
+   * the fetched value is newer than the cached value,
+   * in which case the fetched value is cached and returned wrapped in a CompletionStage.
+   * 
+   * @param key to look up in the cache
+   * @param fetcher called if not value is associated with the key in the cache
+   * @param comparator called if a cached value exists when the fetcher returns 
+   * @return a future containing the cached value or null if not values exist 
+   */
+  CompletionStage<V> get(K key, Function<K, V> fetcher, Comparator<V> comparator);
```
galderz: From an asynchronous perspective, should `Function<K, V>` be `Function<K, CompletionStage<V>` instead? 

The problem then is how the comparator, you'd probably need a custom comparator that has something like this?

    CompletionStage<Integer> compare(CompletionStage<T> o1, CompletionStage<T> o2)

wburns: The comparator should still work as is, we would just have to do use `CompletionStage#thenCombine` internally.

TrueCacheAPI.md
wburns: 👍

wburns: We could provide even a blocking and non blocking variant.

cescoffier: Do you want to redeem `null` or use `Optional`?

TrueCacheAPI.md
cescoffier: @wburns Yes.

TrueCacheAPI.md
galderz: > We could provide even a blocking and non blocking variant.

Not sure about providing a blocking variant since it can be derived from the non blocking one? A blocking one can be created out of taking `V` and converting it to `CompletionStage<V>`, but I see that more like a static helper method or something. I want to avoid overloading.

galderz: > Do you want to redeem null or use Optional?

Eventually should be Optional. So this would eventually look like this?

```
CompletionStage<Optional<V>> get(K key, Function<K, CompletionStage<Optional<V>> fetcher);
```

TrueCacheAPI.md
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Remove entry from the cache and remove it from source by calling the remover function.
+   * 
+   * @param key to delete from cache and source
+   * @param remover function that deletes data from source
+   * @return a future that is completed when key has been removed from both cache and source  
+   */
+  CompletionStage<Void> remove(K key, Consumer<K> remover);
```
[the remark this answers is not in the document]
galderz: The point of the `Consumer<K>` is that it defines how the key is removed in the source of truth (e.g. a database). Just like the `get()`, we want the cache to have some control over this so that it can easily know when the source of truth has completed. 

The problem with your suggestion is that the cache cannot know when the source of truth remove has completed.

The code above is blocking, so to be fully non-blocking it should look more like this:

```
CompletionStage<Void> remove(K key, Function<K, CompletionStage<Void>> remover);
```

PR #15, opened 2019-12-13
Data Store

tristantarrant:


DataStore.adoc
```diff
@@ -0,0 +1,28 @@
...
+Every time a stable topology is installed in the cluster, it should be persisted to the global state. Currently this only happens on shutdown.
+
+=== Restarting nodes
+
+When a node is restarted it should read the topology stored in the global state and join only if it matches the stable topology of the existing cluster.
+
+=== Open issues
+
+When the segment owners are missing, an application will not be able to read/write any data belonging to those segments even if it is "new" data (i.e. keys that were not in those segments when they disappeared).
+We could create temporary segments on the remaining nodes which would then be merged back into the real owners using the conflict resolution logic.
```
ryanemerson: How about if we allow "temprorary segments" when `when-split=allow-read-writes` and throw a `AvailabilityException` when `when-split=deny-read-writes`. `allow-reads` would be the same as it currently is, i.e. just allow reads if a segment node is missing, but allow reads.

danberindei:
One open question for me is what happens in Kubernetes when a pod is stopped permanently, when the admin reduces the number of replicas. Can we get that pod's persistent volume back, attached to the next pod that is started up, or is is going to be lost forever?

DataStore.adoc
ryanemerson: The use of "temporary segments" should probably be configurable, because in some circumstances it might not be possible to resolve the conflicts deterministically.

ryanemerson:
> One open question for me is what happens in Kubernetes when a pod is stopped permanently, when the admin reduces the number of replicas. Can we get that pod's persistent volume back, attached to the next pod that is started up, or is is going to be lost forever?

@danberindei StatefulSets is a hard requirement for Infinispan, so this ensures that the volumes are not deleted automatically. SS volumes must be explicitly removed by admin. https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#limitations

DataStore.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+* ALWAYS: rebalancing happens both on nodes joining and leaving the cluster.
+* SCALE_UP: rebalancing only happens when nodes join.
+* NEVER: rebalancing is disabled in all situations.
+
+Every time a stable topology is installed in the cluster, it should be persisted to the global state. Currently this only happens on shutdown.
+
+=== Restarting nodes
+
+When a node is restarted it should read the topology stored in the global state and join only if it matches the stable topology of the existing cluster.
```
wburns: How would a node know it is restarted? Is it by checking if it already has a global state?

danberindei:
Thanks for the link @ryanemerson! It's pretty clear from their description the volumes are reused when the number of replicas is decreased and then increased back, so it's safe to disable rebalancing on scaling down.

Looks like we also get a grace period between getting the SIGTERM signal and SIGKILL, and the server could register a shutdown hook to send a LEAVE request to the coordinator. Assuming a big enough timeout, we could even delay the shutdown to wait for the other nodes to rebalance, but of course there's no way to enforce a big enough timeout, and if multiple nodes are being shut down in sequence, we'd go in DEGRADED mode anyway. 

Regardless of whether we send a LEAVE command or not, there's a complication that won't allow us to use the latest stable consistent hash with scaling back up: the cache only gets DEGRADED after losing `numOwners` nodes. That means the cache stays AVAILABLE and accepts writes for some time with only 1 owner for some of the segments, and on scale up we'll know which node was the last owner of those segments in order to discard (or merge) the data from the other owners.

DataStore.adoc
danberindei: Yes, it would see the consistent hash with it as a member in the global state.

danberindei:
Thanks Tristan!

PR #17, opened 2020-01-28
Cluster backup tool

ryanemerson:
Currently this just outlines the general approach. Specific details about CLI syntax for more advanced/optional features, such as cache specific backup/export, will be added if the initial idea is suitable.

### DIFF
--- a/cluster-backup-tool.md
+++ b/cluster-backup-tool.md
@@ -46,41 +46,48 @@ Content-Transfer-Encoding: binary
 ```
 
 # Archive Format
-A cluster backup consists of a single global manifest which contains clusterwide metadata. As well as `.xml` and `.dat`
-files for each cache and a single `counters.dat` file. These files are then distributed as a archive format, which can
-be passed to the CLI in order to initiate a import.
+A cluster backup consists of a global manifest which contains clusterwide metadata, such as the version. As well as a
+directory structure that contains all cache templates, cache instances and counters. These files are then distributed
+as a archive format, which can be passed to the CLI in order to initiate a import.
 
 ```
+cache-configs/
+cache-configs/some-template.xml
+cache-configs/another-template.xml
 caches/
 caches/example-user-cache.dat
 caches/example-user-cache.xml
 counters/
 counters/counters.dat
 manifest.properties
-infinispn.xml
+cache-container.xml
 ```
 
 The above files will be packaged as a single `.zip` distribution to aid backup/import; this provides compression as well
 as being compatible with both Linux and Windows.
 
-## Global Manifest
+## Manifest
 Basic java properties file with key/values.
 
 ```java
 version=11.0.0.Final
-globalconfiguration=infinispan.xml
+cache-configs="some-template, another-template"
 caches="org.infinispan.CONFIG", "example-user-cache"
 counters="example-counter"
 ```
 
 The Infinispan version can be used by the Server to quickly determine if a migration to the new server version is possible
-based upon the number of major versions we support migrations across. Similarly, the Global configuration can be used to
-determine if the backup depends on any additional user classes, e.g. serialization marshallers, and can check that the
-server contains these resources on it's classpath, failing fast if they are not present.
+based upon the number of major versions we support migrations across.
 
+## Container XML
+The container XML can be used to determine if the backup depends on any additional user classes, e.g. serialization marshallers,
+and can check that the server contains these resources on it's classpath, failing fast if they are not present.
+
+## Template files
+Each defined template is represented by a `<template-name>.xml` file.
 
 ## Cache Files
-A cache backup consists of a `.xml` and `.dat` file, each of which is named after the cache. The `.xml` file contains
+A cache backup consists of a `<cache-name>.xml` and a `<cache-name>.dat` file. The `.xml` file contains
 the cache configuration and is used to create the initial empty cache. The `.dat` file is a batch file containing the cache
 content and is read by the server in order to restore entries to the cache. Each line in the file corresponds to a cache
 entry, with keys and values separated by a single space.
--- a/cluster-backup-tool.md
+++ b/cluster-backup-tool.md
@@ -28,117 +28,113 @@ Cluster Backup Tool
     
 
 # Architecture
-The backup tool can be implemented as an extension to the CLI, with existing cache data stored as the corresponding CLI
-commands required to add the data to a cache. A cluster backup consists of a single global manifest which contains clusterwide
-metadata. As well as `.json` and `.dat` files for each cache and a single `counters.dat` file. These files are then
-distributed as a archive format, which can be passed to the CLI in order to initiate a import.
+The backup tool should be implemented on the server and be accessible via the `/v2/cluster` REST endpoint, allowing the various
+Infinispan clients (CLI, Console, Operator) to utilise it. Backups can be initiated via a GET call and restored by
+uploading an archive via a POST.
+
+Backup:
+```
+GET /v2/cluster?action=backup
+```
+
+Restore:
+```
+POST /v2/cluster?action=restore
+Content-Disposition: form-data; name=`"file`"; filename=`"backup.zip`"
+Content-Type: application/zip
+Content-Transfer-Encoding: binary
+```
+
+# Archive Format
+A cluster backup consists of a single global manifest which contains clusterwide metadata. As well as `.xml` and `.dat`
+files for each cache and a single `counters.dat` file. These files are then distributed as a archive format, which can
+be passed to the CLI in order to initiate a import.
+
+```
+caches/
+caches/example-user-cache.dat
+caches/example-user-cache.xml
+counters/
+counters/counters.dat
+manifest.properties
+infinispn.xml
+```
+
+The above files will be packaged as a single `.zip` distribution to aid backup/import; this provides compression as well
+as being compatible with both Linux and Windows.
+
+## Global Manifest
+Basic java properties file with key/values.
+
+```java
+version=11.0.0.Final
+globalconfiguration=infinispan.xml
+caches="org.infinispan.CONFIG", "example-user-cache"
+counters="example-counter"
+```
+
+The Infinispan version can be used by the Server to quickly determine if a migration to the new server version is possible
+based upon the number of major versions we support migrations across. Similarly, the Global configuration can be used to
+determine if the backup depends on any additional user classes, e.g. serialization marshallers, and can check that the
+server contains these resources on it's classpath, failing fast if they are not present.
+
+
+## Cache Files
+A cache backup consists of a `.xml` and `.dat` file, each of which is named after the cache. The `.xml` file contains
+the cache configuration and is used to create the initial empty cache. The `.dat` file is a batch file containing the cache
+content and is read by the server in order to restore entries to the cache. Each line in the file corresponds to a cache
+entry, with keys and values separated by a single space.
+
+> If the cache is empty during the backup, then no `.dat` file is created.
+
+Example  `.dat` content:
+```java
+hello world
+hola mundo
+```
+
+> How to represent metadata and handle expiration?
+
+### Storage MediaTypes
+If a cache has a configured key and/or value MediaType, that can be readily represented as a string, e.g. text/plain or json,
+then the key/value should be represented in the `.dat` file using that format. Otherwise a binary MediaType, such
+as `application/x-protostream`, are represented as a Base64 String.
+
+Example  `.dat` content:
+```java
+hello V29ybGQ=
+```
+
+### Internal Caches
+Volatile internal caches should not be included in a backup. Internal caches can be excluded during the creation of the
+backup archive, with no `.xml` or `.dat` files created.
+
+The counter cache should also be omitted in favour of a dedicated `counters.dat` file.
+
+## Counters File
+The `counters.dat` file that contains the informated required in order to recreate all counters and their values
+at backup time. If no counters exist, this file can be omitted from the archive.
+
+```
+<NAME> <STORAGE> <TYPE> <INITIAL_VALUE> <CURRENT_VALUE>
+example-cache PERSISTENT STRONG 0 5
+```
+
+# CLI Integration
+The CLI will be the first client to expose the backup/restore capabilities, proposed syntax:
 
 Backup:
 ```bash
 bin/cli.sh -c localhost:11222 --backup <archive-file>
 ```
 
-Import:
+Restore:
 ```bash
 bin/cli.sh -c localhost:11222 --restore <archive-file>
 ```
 
-## Global Manifest
-Contents:
-- Infinispan Version used to create backup
-- Global configuration
-- Cache names
-
-The Infinispan version can be used by the CLI to quickly determine if a migration to the new server version is possible
-based upon the number of major versions we support migrations across. Similarly, the Global configuration can be used to
-determine if the backup depends on any additional user classes, e.g. serialization marshallers, and can check that the
-server contains these resources on it's classpath, failing fast if they are not present. Finally, the cache metadata is a
-map of all cache names and their associated import file that are contained in the backup.
-
-> A key assumption is that in order to import a data backup to Infinispan version `x`, it's necessary for version `x` of
-the CLI to also be used.
-
-### Format
-Basic json file with key/values, allowing for future changes to ordering and for the GlobalConfiguration json to be
-nested.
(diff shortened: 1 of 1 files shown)


cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+### Internal Caches
+Volatile internal caches should not be included in a backup. Internal caches can be excluded during the creation of the
+backup archive, with no `.json` or `.dat` files created for the blacklisted caches.
+
+The counter cache should also be omitted as CLI import statements in the `counters.dat` file should be used instead. 
+
+### Storage MediaTypes
+If a cache has a configured key and/or value MediaType, that can be readily represented as a string, e.g. text/plain or json,
+then the `cache put` batch commands should utilise that encoding. Otherwise a binary MediaType, such
+as `application/octet-stream` is used and the key/value represented as Base64.
```
wburns, line 102: Why bother changing to Base64? Is this so it is easier for the cli to consume?

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+at backup time. If no counters exist, this file can be omitted from the archive.
+
+```
+create counter --initial-value=<value> --storage=<PERSISTENT|VOLATILE> --type=<STRONG|WEAK> <name>
+```
+
+## Backup Archive Format
+The backup files should be packaged as a single archive distribution to aid backup/import.
+
+> Is `.zip` sufficient?
```
wburns, line 126: I would say so for our Windows bretheren.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+
+> Should we make this optional to allow for greater flexiblity. Default to directory backup? Optional --compress flag?
```
wburns, line 128: Doesn't the restore command the top take a single file? I think having them all separate would be a bit more arduous. Especially as you may end up copying this to another machine.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+
+## CLI
+* Encoding:
+    - Ability to specify --key-encoding and --value-encoding separately
+        - Ideally it would be possible to specify once per batch, with subsequent puts just requiring `put <key> <value>`
+    - Enable non text/plain values to be put inline
+        - e.g `cache put --key-encoding=application/octet-stream --value-encoding=application/octet-stream a2V5Cg== c29tZSB2YWx1ZQ==`
+
+## Rest
+* Ability to iterate over cache entries
```
wburns, line 141: So it wasn't quite clear but how is the backup generated? Is it that the server responds with the values to the CLI and it writes the file? Or is it that the server generates a file on its local disk and then sends it back?

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+    - Recreate data from archive
+- Must be backwards Compatible across x major versions
+- Supported in client/server mode
+
+## Optional/Future
+- Topology aware restoration
+    - Disabled by default, focus is on data
+
+- Import failure modes:
+    - Clusterwide: Rollback all on first error
```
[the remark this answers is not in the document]
ryanemerson: My assumption here was that the cluster importing the archive was new, so the caches would be deleted on failure. If the importing cluster already contains a cache with the equivalent name, the import should probably fail. Rollingback only the imported cache entries would be more difficult and we would probably have to leverage the `CommitManager`.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Cache specific imports
+    - Greater flexibility
+    - Ability to retry failed cache imports
+    - Convert from one cache type to another
+
+- Configuration only backup/import
+
+- Enable "online" backups where the cluster creates a snapshot from a live cluster
+    - Utilise `CommitManager` as used by state transfer and conflict resolution
```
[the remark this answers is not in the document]
ryanemerson: If the backup is possible on a "online" cluster the idea is that the `CommitManager` is used to track all of the entries that are put after the backup has been initiated, but before it has completed, as there has to be a cut-off point for when cache entries are accepted as part of the backup. Otherwise in a heavily loaded cache it's possible the backup will never complete.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+    
+
+# Architecture
+The backup tool can be implemented as an extension to the CLI, with existing cache data stored as the corresponding CLI
+commands required to add the data to a cache. A cluster backup consists of a single global manifest which contains clusterwide
+metadata. As well as `.json` and `.dat` files for each cache and a single `counters.dat` file. These files are then
```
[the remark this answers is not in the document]
ryanemerson: I think so as counters are created via the CLI as follows:

```
create counter --initial-value=<val> --storage=<PERSISTENT|Volatile> --type=<strong|weak> <counter-name>
```
https://infinispan.org/docs/stable/titles/cli/cli.html#creating_counters

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+- Global configuration
+- Cache names
+
+The Infinispan version can be used by the CLI to quickly determine if a migration to the new server version is possible
+based upon the number of major versions we support migrations across. Similarly, the Global configuration can be used to
+determine if the backup depends on any additional user classes, e.g. serialization marshallers, and can check that the
+server contains these resources on it's classpath, failing fast if they are not present. Finally, the cache metadata is a
+map of all cache names and their associated import file that are contained in the backup.
+
+> A key assumption is that in order to import a data backup to Infinispan version `x`, it's necessary for version `x` of
```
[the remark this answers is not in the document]
ryanemerson: Say you're importing an archive from Infinispan 10 to Infinispan 11. If you enforce that the CLI of 11 has to be used, then any transformations required for compatibility reasons only have to be present on 11. If a user is deploying an Infinispan 11 cluster, I don't see why they wouldn't also have access to the 11 CLI.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+the CLI to also be used.
+
+### Format
+Basic json file with key/values, allowing for future changes to ordering and for the GlobalConfiguration json to be
+nested.
+```json
+{
+	"version": "10.1.1.Final",
+	"global-configuration": "...",
```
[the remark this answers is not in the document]
ryanemerson: It's probably worth recording all of the global configuration, however things like `globalState.persistentLocation` won't have an impact during the import phase as the import will be performed against an already running cluster that has these values configured before startup.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Cache Files
+A cache backup consists of a `.json` and `.dat` file, each of which is named after the cache. The `.json` file is a json
+representation of the cache's configuration and the `.dat` file is the batch file executed by the CLI as part of the import.
+Below is an example `.dat` file for a cache with the name ``example-user-cache`:
+
+1. Create the cache using the `.json` file:
+
+```
+create cache --file=example-user-cache.json example-user-cache
```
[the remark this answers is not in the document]
ryanemerson: In effect it's a batch script which is executed sequentially by the CLI. The user shouldn't have todo any manual steps other than initiating the import via the cli and providing the archive tar.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
...
+```
+cd caches/example-user-cache
+put hello world
+put hola mundo
+```
+> If the cache is empty during the backup, then no `.dat` file is created.
+
+### Internal Caches
+Volatile internal caches should not be included in a backup. Internal caches can be excluded during the creation of the
+backup archive, with no `.json` or `.dat` files created for the blacklisted caches.
```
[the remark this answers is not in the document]
ryanemerson: The protobuf cache is persisted. Are there internal caches that are user facing and volatile?

cluster-backup-tool.md
ryanemerson: Yeah that was my thinking.

cluster-backup-tool.md
ryanemerson: Good point :+1: 

I'll look at how Tristan has handled  this for the server report.

cluster-backup-tool.md
ryanemerson: Agreed. I don't see any real use-case for just creating a directory.

cluster-backup-tool.md
ryanemerson: Originally I was thinking on the server, however I'm now thinking that the CLI retrieving entries would be more flexible as it would allow us to failover and continue the backup in the event of a server failure.

cluster-backup-tool.md
```diff
@@ -0,0 +1,144 @@
+Cluster Backup Tool
+====================
+# Requirements
+## Hard
+- Ability to:
+    - Backup data to a file(s) at a given point in time
```
pruivo, line 6: any idea how to achieve this?

cluster-backup-tool.md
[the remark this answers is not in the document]
pruivo: I would "ignore" the volatile counters. 
And you will need a file to store the counter's configuration.
One possible issue it the CLI itself. It doesn't have an operation to set the counter's value. And weak counters can only do increments/decrements.

cluster-backup-tool.md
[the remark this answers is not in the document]
pruivo: so why store the GlobalConfiguration? If the server is already running, can you change its configuration?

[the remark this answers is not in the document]
ryanemerson: Certain aspects you can, for example the marshalling configuration. Of course we could just store what we know to be configurable at runtime, however I think we should store the entire global configuration for the following reasons:

1) In a future major version what is configurable at runtime may change.
2) Storing a subset can only ever be more restrictive than storing the entire config.
3) The difference in byte stored between storing a subset and the entire global configuration is negligible and reducing a cluster acrchive by a few bytes is not a priority.

cluster-backup-tool.md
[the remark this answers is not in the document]
ryanemerson: > I would "ignore" the volatile counters.

I disagree as the purpose of the tool is to create a backup of the clusters current state so that it can be restored at a later date. We include non-persistent caches, so I don't see why we would do the same for counters?

> And you will need a file to store the counter's configuration.

That's the purpose of the `counters.dat` file, it's a batch file that recreates all of the counters with their initial configuration and current value.


> One possible issue it the CLI itself. It doesn't have an operation to set the counter's value. And weak counters can only do increments/decrements.

I was thinking that we can calculate the delta between the current value and the initial-value and then call:

```
[//containers/default/counters/my-weak-counter]> add --delta=2
```

It could also be possible to add `--delta` to the `create counter` command so that this can be achieved on a single line.

cluster-backup-tool.md
ryanemerson: I'm not sure I understand the question. This whole document is essentially a proposal on how to do ^.

pruivo: so, transactional caches cross my mind, more concretely, how to achieve a valid snapshot.
Since tx involves multiple [primary] owners, iterating over the keys may not be a good idea. And starting a cluster from backup with some partial transactions committed may not be a good idea.

cluster-backup-tool.md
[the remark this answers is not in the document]
pruivo: > I disagree as the purpose of the tool is to create a backup of the clusters current state so that it can be restored at a later date. We include non-persistent caches, so I don't see why we would do the same for counters?

in my head, if a user configures a volatile counter, it is expected to be lost in case of shutdown, restarts, or crashes. 

I guess it wouldn't hurt to have them included :+1:

cluster-backup-tool.md
ryanemerson: So the initial scope of this tool is to create a backup from an "offline" cluster, i.e. one that is not currently serving requests, so in this case there won't be any pending transactions. If one or more transactions are pending on a cache, we could potentially wait for them to complete before proceeding.

Eventually we will want "online" backups of a live cluster, which will definitely have to take into account transactional boundarys and defining what a valid snapshot is. This is non-trivial and realistically will probably be best suited to a breakout session at our next F2F (whenever that will be). I make reference to this in the `Optional/Future` section here: https://github.com/infinispan/infinispan-designs/pull/17/files#r418168778

cluster-backup-tool.md
```diff
@@ -0,0 +1,164 @@
...
+and can check that the server contains these resources on it's classpath, failing fast if they are not present.
+
+### Template files
+Each defined template is represented by a `<template-name>.xml` file.
+
+### Cache Files
+A cache backup consists of a `<cache-name>.xml` and a `<cache-name>.dat` file. The `.xml` file contains
+the cache configuration and is used to create the initial empty cache. The `.dat` file is a batch file containing the cache
+content and is read by the server in order to restore entries to the cache. Each line in the file corresponds to a cache
+entry, with keys and values separated by a single space.
```
rvansa, line 110: I guess the serialization format of binary keys and values is a bit more complex than that. Using separators (in this case space and newline) gets you into the escaping business. And what about the metadata?

cluster-backup-tool.md
```diff
@@ -0,0 +1,164 @@
...
+Example  `.dat` content:
+```java
+hello world
+hola mundo
+```
+
+> How to represent metadata and handle expiration?
+
+#### Storage MediaTypes
+If a cache has a configured key and/or value MediaType, that can be readily represented as a string, e.g. text/plain or json,
```
rvansa, line 123: Should keys with spaces and newlines be base64-encoded, too?
Are metadata dropped completely?

rvansa: I just wonder what is the advantage of storing in text format rather than binary. While zipping mostly helps with base64 introduced inflation, I don't see much 'pros'.
```
$ cat /dev/urandom | head -c 65536 > /tmp/r
$ base64 /tmp/r > /tmp/b
$ zip /tmp/b.zip /tmp/b
$ zip /tmp/r.zip /tmp/r
$ ls -l /tmp/b* /tmp/r*
-rw-rw-r--. 1 rvansa rvansa 88534 Jun 19 10:42 /tmp/b
-rw-rw-r--. 1 rvansa rvansa 67444 Jun 19 10:43 /tmp/b.zip
-rw-rw-r--. 1 rvansa rvansa 65536 Jun 19 10:42 /tmp/r
-rw-rw-r--. 1 rvansa rvansa 65706 Jun 19 10:43 /tmp/r.zip
```

ryanemerson: > You could not tell when it's plaintext and when it's base64...

The plan originally was to rely on the cache configuration to determine this.


> Are metadata dropped completely?

No, I mentioned it above but forgot to detail it's representation. The metadata will have to be serialized and represented accordingly. The open question I have myself is how to handle expiration and max-idle? On backup restore, do we just set the entry creation and last used times to the time of the restore.

> I just wonder what is the advantage of storing in text format rather than binary.

Thanks for mentioning this ... there is none AFAICT. This is a hangover from my initial design where I was going to piggyback on the CLI syntax. Storing in binary format is the way to go :+1:

rvansa: > The open question I have myself is how to handle expiration and max-idle? On backup restore, do we just set the entry creation and last used times to the time of the restore.

There's already a precedent in stores, isn't it? (I just don't remember what that is) What you suggest is simpler but obviously wrong - you should store these timestamps in metadata as is and not change then during restore.

ryanemerson: > There's already a precedent in stores, isn't it?

Yes, the original values are respected.

> you should store these timestamps in metadata as is and not change then during restore

I'm imagining a scenario where a user has created a backup of a cache, totalling ~ 1Gb. They then restore the backup and have 0 entries in the cache because they have all expired. We could make this behaviour configurable during the import by adding a param, e.g. `/v2/sever?action=import&resetTimestamps=true`.

rvansa: A wise man (more than) once said "less knobs".

If the data is set with expiration, it was set so for a reason - it's not relevant anymore or it even should not be accessible (e.g. here as part as Covid tracking you can give the health officer access to your cell phone location history; all data should be wiped out shortly after the session, though). Cluster failure & backup restore is a technical detail that does nothing to do with the data relevancy.

wburns: I agree, if it has expiration metadata and is now expired - just don't reinstate that data.

PR #18, opened 2020-02-18
ISPN-11304 Scaling without state transfer

danberindei:
https://issues.redhat.com/browse/ISPN-11304


gustavonalle:
From the design document, it seems there are many limitations in a cluster scaled like this (no indexing/query, no data redundancy, has data loss, extra side car cache with all the keys), apart from the fact it introduces more config (aka extra knob) and extra complexity. 
What is the use case? It is not described in the JIRA

tristantarrant:
@gustavonalle this is a requirement of the https://cloudbutton.eu/ project: they need to be able to scale linearly, adding more capacity, without the impact of state transfer. The data only needs to be kept until the end of the processing, after which it is dumped elsewhere.

scaling-without-state-transfer.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+can replace the invalidation interceptor with a custom interceptor.
+
+The interceptor will first look up affected keys in the anchor cache.
+If the key doesn't exist in the anchor cache, reads can assume the value is missing.
+Writes must insert the current writer's location in the anchor cache,
+then forward to the current writer for the actual write.
+
+When a key is removed, the interceptor must also remove the key from the anchor cache.
+When a node leaves, the interceptor must remove all the keys mapped to that node
+from the anchor cache.
```
rvansa: I guess that since you don't expect the cache to be reliable (no backups), so here you don't care about concurrency (adding and removing key at the same time might result in anchor cache and the data cache not in sync).
However that also means that you might start leaking entries in the data cache without anchor cache knowing - do you intend to avoid garbage or prune it periodically?

rvansa:
@tristantarrant Do they need to provide keys themselves, or could they just get a 'handle' from Infinispan? Because if it's the latter it might make more sense to disable state transfer and give them keys that contain the address internally (and adjust the hashing function to work with those keys).

scaling-without-state-transfer.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+
+== Configuration
+
+Configuration means a custom element `anchored-keys`
+with a single attribute `enabled`.
+
+```
+<invalidation-cache name="name">
+    <anchored-keys enabled="true"/>
```
ryanemerson: Unless you think there's a need for additional configuration, I think it would be simpler to have this as a boolean attribute on the invalidation-cache element.

```xml
<invalidation-cache name="name" anchored-keys="true" />
```

scaling-without-state-transfer.asciidoc
danberindei: No, I do care about concurrency, I just haven't given removals much thought yet.

I'm thinking the interceptor will write a sentinel value in the anchor cache on removal, and a periodic task will prune the sentinel values from the replicated cache. Perhaps setting a lifespan on the sentinel entries would also work.

Writes will have more work to do, but hopefully I can still avoid `compute`, which doesn't yet work with off-heap.

scaling-without-state-transfer.asciidoc
danberindei: I'd rather put this in a separate module, not in the core, and a module configuration parser needs a root element.

ryanemerson: :+1: makes sense. I did see you mentioned ModuleLifecycle, but I didn't think of the impact on the config.

danberindei:
@rvansa FWIW the initial plan was to use the replicated cache in a custom `KeyPartitioner`, mapping keys to segments dynamically. The main problem with that is that the `KeyPartitioner` doesn't know whether it's computing a segment for reading or for writing, so it will create an entry in the anchor cache for all missing reads.

I also implemented a custom `SparseConsistentHash` with most segments assigned to a placeholder address, and that placeholder CH member created its own share of problems. In retrospect, it would have been easier to assign all segments to the current writer and to prevent remote gets to the current writer for missing keys from a custom interceptor.

Indeed, if we generate the keys ourselves we wouldn't need a side cache for the key locations. I even fantasized for a bit we could add support for custom `KeyPartitioner`s on the client and make the key->segment deterministic in order for the client requests to go directly to the owner, but I don't think it would work. Topology views sent to the client must include the member index for each segment, so the number of segments we can use is limited (I used `Short.MAX_VALUE`), and the birthday paradox ensures that we'd have collisions very soon.

rvansa:
@danberindei I think that in this case thinking of segments is already too far towards the current model while you're trying to do something completely different. I would stub everything with single segment; on client side there's `org.infinispan.client.hotrod.impl.consistenthash.ConsistentHash.getServer(Object)` which would know how to parse the concrete IP from the key. On the core side, I forgot that most of the code now uses segment ids stored in the commands instead of hashing the key again and again. That would probably require bypassing the distribution logic on quite a few places. Unless you'd just disable also the client retry logic (or rather force it to retry on the same server) and read/write the entry wherever the request lands, moving all intelligence to client only.

danberindei:
Thanks for reminding me that the `ConsistentHash` implementation is configurable in the client @rvansa , I kept thinking that we'd need to add support for custom `KeyPartitioners`!

Moving the distribution logic to the client should work, assuming the client is the one generating the keys on behalf of the application and the application doesn't need embedded access. Making bulk operations and listeners work would likely require big changes in the client as well. /cc @tristantarrant 

Making the embedded distribution logic know about server IPs would add another level of complexity, so I'd rather avoid it. The `FailoverRequestBalancingStrategy` is also configurable on the client, but we don't know which key we're retrying for, so it's probably best to disable retrying.

rvansa:
@danberindei You don't know which key you're retrying for, but you know which server - `nextServer` has those failed in a list - just pick it. Therefore you can just return the same server over and over.

PR #20, opened 2020-09-23
Kubernetes CLI

tristantarrant:


Kubernetes-CLI.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+Installing the executable as `kubectl-infinispan` allows invocation as a `kubectl/oc` plugin, e.g.:
+
+`kubectl infinispan ...`
+
+Using the Fabric8 Kubernetes Client we can automatically detect the existence of a Kubernetes configuration (default to `~/.kube/config`) and automatically connect to the Kubernetes cluster.
+
+=== Kubernetes commands
+
+The Infinispan Kubernetes CLI plugin defaults to using the `datagrid` namespace and the `infinispan` service name. These can be overridden in each command using the `--namespace` parameter and the name argument.
```
Crumby, line 20: Default namespace should default to current context (can be retrieved eg by `oc project` or found in config). If the plugin will default to different namespace it will create a confusion. `kubectl/oc` would operate on top of different namespaces depending on whether user is using core `oc/kubectl` or operating with the plugin.

Kubernetes-CLI.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+==== Scale an Infinispan service
+
+`kubectl infinispan scale -r replicas [servicename]`
+
+Scale an Infinispan service to the specified number of replicas
+
+===== Interact with a running Infinispan service
+
+`kubectl shell servicename`
```
Crumby, line 48: Is `infinispan` dropped intentionally in here? Won't this require extra kubectl-shell plugin to be developed? Since it's supposed to be infinispan specific I think it should be "namespaced": `kubectl infinispan shell(connect?) servicename`

Kubernetes-CLI.adoc
```diff
@@ -0,0 +1,63 @@
...
+==== Install the operator
+
+`kubectl infinispan install [-n namespace]`
+
+Install the Infinispan Operator in the specified namespace. 
+This should also install the relevant OLM subscription and operator group so that the operator can be updated.
+
+==== Create an Infinispan service
+
+`kubectl infinispan create service [-n namespace] [-r replicas] [--expose-type=type] [--expose-port=port] [--expose-host=hostname] [servicename]`
```
Crumby, line 31: What type of service will this command create, cache or datagrid? The default by operator is `cache` but that has pretty much limited capabilities. Maybe having `[-t serviceType]` option here wouldn't be bad.

Also I'm trying to think whether `service` is a good verb in here. Even though I get why it was picked up, it pretty much fights with standard OpenShift `service` concept (and we might want it for potential service customization in the future?). Have you considered `cluster` or `server`? As `Infinispan Cluster` is used for referencing infinispan resources in the console there would be some consistency.

Other options could be `kubectl infinispan create infinispan` or not using a verb at all. Former doesn't feel exactly catchy and the later will probably cause issues maintaining different CRDs.

Crumby:
Looks like a solid base. One from many another extensions of this concept could be setting auth and encryption secrets eg.: `kubectl/oc infinispan set authSecret serviceName [--secret secretName|--file identities.yaml]`. But that's probably out of scope for Infinispan 12?

Kubernetes-CLI.adoc
Crumby: Since it doesn't allow auth specification it'd be awesome if the create command would output the login details.

Kubernetes-CLI.adoc
tristantarrant: I'll amend that.

Kubernetes-CLI.adoc
tristantarrant: Oversight.

Kubernetes-CLI.adoc
tristantarrant: I'll use cluster. Caches will be created using `kubectl infinispan create cache`

oraNod:
Would it be possible to have commands for things like verifying clusters are well formed? 

e.g. instead of `kubectl get infinispan -o yaml` something link `kubectl infinispan show [clusterview|xsiteview]`

Kubernetes-CLI.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+Using Quarkus we can build the CLI as a native executable for the three most common platforms: Linux x86_64, OS X, Windows.
+
+== Kubernetes mode
+
+Installing the executable as `kubectl-infinispan` allows invocation as a `kubectl/oc` plugin, e.g.:
+
+`kubectl infinispan ...`
+
+Using the Fabric8 Kubernetes Client we can automatically detect the existence of a Kubernetes configuration (default to `~/.kube/config`) and automatically connect to the Kubernetes cluster.
```
danberindei, line 16: If `~/.kube/config` is only the default, is there be a parameter or environment variable that can make the CLI use a different configuration?

Crumby: I believe it should work out of the box with `KUBECONFIG`

PR #21, opened 2020-11-03
Create operator_rolling_upgrades.md

ryanemerson:


operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+metadata:
+  name: example-infinispan
+spec:
+  upgrades: rolling
+```
+
+## Procedure
+When an upgrade is possible, the operator performs the following for each Infinispan CR:
+
+1. Create a new StatefulSet for the target cluster based upon the CR spec, wait until the cluster is formed
```
ryanemerson, line 19: Is it Ok to create a new `-ping` service for the new Statefulset that utilises a different label value? This will ensure that the two clusters don't merge.

Crumby: I believe this will be necessary. Shouldn't cause any issues.

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+<remote-store xmlns="urn:infinispan:config:store:remote:12.0"
+                 cache="<cache-name>" 
+                 protocol-version="<protocol-version>"> 
+      <remote-server host="<ip-of-pod>" port="11222"/> 
+   </remote-store>
+```
+
+4. Switch clients to the target cluster.
+
+5. Update the CR spec.StatefulSetName to new Statefulset
```
Crumby, line 35: There is no `spec.StatefulSetName` only `status.StatefulSet...` one. What would be the purpose of this step?

ryanemerson: :+1:  I meant to say `status.StatefulSet`

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+
+```xml
+<remote-store xmlns="urn:infinispan:config:store:remote:12.0"
+                 cache="<cache-name>" 
+                 protocol-version="<protocol-version>"> 
+      <remote-server host="<ip-of-pod>" port="11222"/> 
+   </remote-store>
+```
+
+4. Switch clients to the target cluster.
```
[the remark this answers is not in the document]
ryanemerson: Exactly. Any requests on the target cluster can be serviced via the source cluster due to the remote-store.

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+
+7. Disconnect remote-store on the target cluster for all caches
+```
+POST /v2/caches/{cacheName}?action=disconnect-source
+```
+
+8. Delete the source StatefulSet
+
+## Switching Clients
+- How do we ensure that the single-port service only points to the target cluster?
```
Crumby, line 51: Service targets pods created by `StatefulSet`, you'll change the selector of the service to match only pods of the new `StatefulSet`.

Eg.: Old StatefulSet would be named `infinispan-11.0.1` and the new one `infinispan-12.0.0`. That would be reflected in the pod labels. Thus making it simple to correctly reference by new ping service and update of the `single-port` service

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+## Procedure
+When an upgrade is possible, the operator performs the following for each Infinispan CR:
+
+1. Create a new StatefulSet for the target cluster based upon the CR spec, wait until the cluster is formed
+
+2. Retrieve all cache configurations from the cluster in the original Statefulset
+
+3. Append a remote-store configuration to each of the ^ configurations and create cache target cluster:
```
dmvolod, line 23: @ryanemerson if we need separate configurations for source and target clusters? If yes, the ConfigMap's should be separate also. Or we need to embed this config on the cluster startup.

operator_rolling_upgrades.md
dmvolod, line 35:
```suggestion
5. Update the CR spec.status.StatefulSetName to new Statefulset
```
We should also make StatefulSetName required and provide this filed on the reconcile ASAP

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
+# Operator Rolling Upgrades
+Optionally allow a Infinispan CR cluster to be upgraded with zero downtime and no loss of data.
+
+## Spec
+Add a field to the CR spec that determines whether an upgrade should be `rolling` or `gracefulShutdown`. The latter is the current approach that does not allow data migration.
```
Crumby, line 5: If the user picks `rolling`, will it roll via **procedure** described below for every release or would it be achievable for some of them to be done through in-place (eg. micro from `11.0.1` to `11.0.2`, security fixes to underlying images that won't cause version change at all)?

If the user decides to begin update process from eg 11.0.0 to 12.0.0 and there are 3 more releases to 11. The whole process will be repeated 5 times `11.0.0` -> `11.0.1` -> `11.0.2` -> `11.0.3` -> `12.0.0` which could be quite time (and resource) consuming. Question is, if we managed to do at least some of the `11.0.0` -> `11.0.3` inplace, would it save some time and made the process more seamless?

Another possibility of improvement here is start using `skip` field in CSVs to allow direct update from `11.0.0` to `12.0.0`

operator_rolling_upgrades.md
ryanemerson: Is it possible to update the selector of the service with zero downtime? I'm just wondering about HR client failover.

operator_rolling_upgrades.md
ryanemerson: I think using the same config-map should be fine until an operator upgrade results in a new config being required for the server, in which case we would need to create a new ConfigMap.  So I think it makes sense to future proof the initial design and create a new ConfigMap to cover this possibility.

operator_rolling_upgrades.md
ryanemerson: > If the user picks `rolling`, will it roll via **procedure** described below for every release or would it be achievable for some of them to be done through in-place (eg. micro from `11.0.1` to `11.0.2`, security fixes to underlying images that won't cause version change at all)?

It's an interesting idea for changes to the layers below our image. However, I don't think it would be possible for the patch upgrades to be in-place without us having all of the Infinispan core capabilities and tests that are required for the full in-place rolling upgrade design.



> If the user decides to begin update process from eg 11.0.0 to 12.0.0 and there are 3 more releases to 11. The whole process will be repeated 5 times `11.0.0` -> `11.0.1` -> `11.0.2` -> `11.0.3` -> `12.0.0` which could be quite time (and resource) consuming. Question is, if we managed to do at least some of the `11.0.0` -> `11.0.3` inplace, would it save some time and made the process more seamless?

It would definitely save time for the user, but I fear that this would also add considerable complexity to what is designed to be a quick solution to a larger problem, i.e. solving in-place rolling upgrades.



> Another possibility of improvement here is start using `skip` field in CSVs to allow direct update from `11.0.0` to `12.0.0`

Sounds good to me, but I'm lacking knowledge about OLM to know how this works.

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+When an upgrade is possible, the operator performs the following for each Infinispan CR:
+
+1. Create a new StatefulSet for the target cluster based upon the CR spec, wait until the cluster is formed
+
+2. Retrieve all cache configurations from the cluster in the original Statefulset
+
+3. Append a remote-store configuration to each of the ^ configurations and create cache target cluster:
+
+```xml
+<remote-store xmlns="urn:infinispan:config:store:remote:12.0"
```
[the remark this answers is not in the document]
ryanemerson: Yes, but luckily the POST request in step 7 takes care of this :slightly_smiling_face:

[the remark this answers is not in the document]
dmvolod: Not sure, that's will be persistent. Containers are ephemeral and just ConfigMap stores configuration.

operator_rolling_upgrades.md
dmvolod: > > If the user picks `rolling`, will it roll via **procedure** described below for every release or would it be achievable for some of them to be done through in-place (eg. micro from `11.0.1` to `11.0.2`, security fixes to underlying images that won't cause version change at all)?
> 
> It's an interesting idea for changes to the layers below our image. However, I don't think it would be possible for the patch upgrades to be in-place without us having all of the Infinispan core capabilities and tests that are required for the full in-place rolling upgrade design.

We can pretest this combination and store all possible combinations as a graph for upgrade options definition

operator_rolling_upgrades.md
slaskawi: I believe you can just switch the labels once the target cluster is ready. Once you establish a TCP connection from a client to a server, it's point-to-point. The service is only required when you initiate such a connection. 

This means that the switching the clients from the old to the new will happen automatically as soon as you start wiping the old cluster out.

slaskawi: Just thinking on it a bit more - maybe you should switch the labels as soon as you start the upgrade procedure (not at the end of it)?

operator_rolling_upgrades.md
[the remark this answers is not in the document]
ryanemerson: `remote-store` is added to a cache configuration so this is all done at runtime, it does not affect the .yaml that is passed to the server image.

operator_rolling_upgrades.md
ryanemerson: The problem is that infinispan's core is lacking some of the capabilities required to correctly handle in-place upgrades. This is a long term goal, but is beyond the scope of this design document.

operator_rolling_upgrades.md
ryanemerson: We need to wait for the cache's to be defined on the target cluster with the remote-store configuration, we can then switch the labels before proceeding with the sync. Without the cache config on the target cluster, the caches won't exist and it won't be possible to retrieve old values from the source cluster.

operator_rolling_upgrades.md
```diff
@@ -0,0 +1,54 @@
...
+                 cache="<cache-name>" 
+                 protocol-version="<protocol-version>"> 
+      <remote-server host="<ip-of-pod>" port="11222"/> 
+   </remote-store>
+```
+
+4. Switch clients to the target cluster.
+
+5. Update the CR spec.status.StatefulSetName to new Statefulset
```
Crumby, line 35: Status is not part of the spec, it stands on its own
```suggestion
5. Update the CR status.StatefulSetName to new Statefulset
```

operator_rolling_upgrades.md
Crumby: > Is it possible to update the selector of the service without unavailability? I'm just wondering about HR client failover.

It should take effect immediately or at least there shouldn't be a gap when the service isn't connected to either cluster. But I've never tested it.

operator_rolling_upgrades.md
Crumby: DNS query for JGroups is part of the ConfigMap and we will need a different one for a new cluster

operator_rolling_upgrades.md
Crumby: Right, I don't think it makes sense to have some updates on the channel as an in-place and some via RollingUpgrade. It would unnecessarily complicate the code to recognize which is possible and since  in-place are driven by SS it's spec would change all the time.

operator_rolling_upgrades.md
ryanemerson: Very good point :sweat_smile:

rigazilla, line 23: Do we support local caches? They will need to be created node by node?

rigazilla (commented):
Nice and clear description @ryanemerson. IMO it would be great to add some info on how we want to report upgrade progress on the infinispan.status field of the CR. Do we need new status label? New conditions?

operator_rolling_upgrades.md
rigazilla, line 5: I'm the only one confused by the usage of `rolling`? It conflict with the k8s rolling upgrade procedure in my mind

operator_rolling_upgrades.md
ryanemerson: That's a good question. Do local caches even make sense in openshift? As the pods are always accessed via a service or route it's not simple to connect to a specific node.

If we did support upgrading local caches, then the process would be something like:
1. Map the source and target pods, so `source-0 -> target-0, `source-1 -> target-1` etc
2. For each pod
  i. Create the local cache with the remote store, with the store pointing to the corresponding source pod
  ii. Sync the local cache via rest
  iii. Remove the remote-store from the configuration

operator_rolling_upgrades.md
ryanemerson: I think that's fair, it's a very overloaded term.

How about `hotrodMigration`? That fits with the new name for this capability "Hot Rod Migration"

So eventually we would have:

```yaml
spec:
  upgrade: hotrodMigration | inline | gracefulShutdown
```

rigazilla: How about `migration` only, `hotrod` sounds a technical aspect imo. But it's just a minor detail, thery're both ok to me.

ryanemerson:
@rigazilla Rebased to a single commit. Unless there is more feedback, I think we're good to merge

PR #23, opened 2021-01-21
Authorization 2.0

tristantarrant:


Authorization.adoc
```diff
@@ -0,0 +1,71 @@
...
+Add a CREATE permission to `org.infinispan.security.AuthorizationPermission`.
+CREATE allows creation/removal of caches, counters, schemas, scripts.
+CREATE supersedes and deprecates the existing `___schema_manager` and `___script_manager` internal roles.
+
+=== Default roles
+Enabling authorization without supplying any roles would define the following default roles:
+
+* *admin* superuser, allowed to do everything (*ALL*)
+* *application* allowed to perform all read/write ops, but not allowed to create/remove caches, schemas, scripts (*ALL_READ*, *ALL_WRITE*, *LISTEN*, *EXEC*)
+* *builder* allowed to create/remove caches, schemas, scripts (*ALL_READ*, *ALL_WRITE*, *LISTEN*, *EXEC*, *CREATE*)
```
pruivo, line 31: `LIFECYCLE` too?

Authorization.adoc
```diff
@@ -0,0 +1,71 @@
...
+=== Server authorization 
+
+* The default server configurations should enable authorization OOTB for server/container ops.
+* The default role mapper should be the `cluster-role-mapper` which stores principal-to-role mappings in a replicated persistent cache across the cluster
+
+=== Cluster Role Mapper
+
+* if empty, it behaves like the `identity-role-mapper` so that default behaviour doesn't change, i.e. the identity's principals are mapped 1:1 with roles. 
+This means an "admin" user will act with the "admin" role.
+* CLI `grant` and `deny` commands control the association between principals and roles, e.g. `grant myuser builder`
```
pruivo, line 43: what happens to "groups"? how are the groups related to this?

tristantarrant: "Groups" are only a concept of the realm: what the authorization code sees is just a set of principals:
* a user principal
* 0 or more group principals
* an InetAddress principal in case of a remote connection (Hot Rod or HTTP)
* a Certificate principal in case of an SSL connection with client authentication

pruivo: I think I'm messing the role & group concept in my head :/ 
so groups are also mapped to roles? and, in the end, the user's permission is its "private" permission + all group's permissions?

tristantarrant: groups are principals, so principals are mapped to roles
a user's permission is the union of all permissions of all the roles it belongs to

Authorization.adoc
tristantarrant: admin has ALL which implies LIFECYCLE. 
builder cannot stop the server

ryanemerson (approved):

Authorization.adoc
pruivo: so, the "builder" can create but cannot start a cache?

danberindei: You don't actually need LIFECYCLE to start a cache on the server... you just put the configuration in the `org.infinispan.CONFIG` cache, and the ConfigurationManager component creates+starts it.

PR #24, opened 2021-02-02
Indexed Entities

tristantarrant:
Static declaration of indexed entities

https://issues.redhat.com/browse/ISPN-10292

### DIFF
--- a/Indexed Entitites.adoc
+++ b/Indexed Entitites.adoc
@@ -1,0 +1,59 @@
+== Indexed entities
+
+Defining indexed entities in a cache configuration currently requires those entities to be present in the schema cache. For example:
+
+[source,xml]
+----
+<distributed-cache name="cache">
+   <indexing>
+      <indexed-entities>
+         <indexed-entity>entity.proto</indexed-entity>
+      </indexed-entities>
+   </indexing>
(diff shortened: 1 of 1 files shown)


gustavonalle:
-1 from me, I think the premise of the JIRA is not correct:
```
But this is not always ensured as there might be catastropic failures or unclean shutdown which can cause inconsistence or complete loss.
```

The prototobuf files are never lost. Furthermore, it complicates schema handling since schemas can be changed, deleted, added at runtime too.

tristantarrant:
Ignore the premise of the Jira: this is about the impossibility of defining statically defined indexed caches.

Indexed Entitites.adoc
```diff
@@ -0,0 +1,59 @@
...
+    required string surname = 3;
+}
+         ]]></indexed-entity-schema>
+      </indexed-entities>
+   </indexing>
+</distributed-cache>
+----
+
+On cache startup, the cache manager will ensure that the schema is added to the schema cache unless it is already present. 
+It is up to the user to ensure that external resources are identical across all nodes in the cluster.
```
danberindei, line 59: What about schema updates, since the protobuf schema cache is persistent?

PR #25, opened 2021-04-19
SQL Store

tristantarrant:
https://issues.redhat.com/browse/ISPN-11875

### DIFF
--- a/SQL Store.adoc
+++ b/SQL Store.adoc
@@ -1,0 +1,264 @@
+= SQL Store
+
+As a application developer or data grid administrator, I want to preload a cache from a relational database accessed by JDBC - via configuration without writing boilerplate code.
+
+== Architecture
+
+* the store can be either read/write or read-only (in which case it behaves as a loader)
+* the store can work in either single-table mode or in query-mode (via user supplied queries)
+* the store can work with simple keys/values or with complex types defined as protobuf schemas
+* a protobuf schema can either be generated dynamically from the resultset metadata or supplied by the user
+* the key can either be a single column or composite
+* the store is to be implemented as part of the existing JDBC store module so that it can reuse the general configuration of connection pooling/datasources and the dialect helpers
+
+== Modes
+
+=== Table mode
+
+This mode is the simplest, in that it just requires the user to supply a single table name.
+Through the database metadata, the store will determine:
+
+* the column names and types
+* which columns are part of the primary key
+
+The proposed XML configuration for this mode:
+
+[source,xml]
+----
+<local-cache name="books">
+   <persistence>
+      <table-jdbc-store>
+         <data-source jndi-url="..." />
+         <!-- the table element is required -->
+         <table name="books"/>
+         <!-- the schema element is optional. Defaults will be used otherwise -->
+         <schema message-name="Book" key-message-name="BookID" package="library" embedded-key="true"/>
+      </table-jdbc-store>
+   </persistence>
+</local-cache>
+----
+
+Optionally, the key can be embedded in the value.
+
+=== Query mode
+
+In this mode, the user supplies the SQL queries to retrieve the entries and (optionally) to insert/update/delete them.
+
+[cols="1,2,1"]
+|===
+| *Query*
+| *Description*
+| *Example*
+
+| *SELECT ALL*
+| The query to use when bulk-loading data into the cache. Columns *MUST* be explicitly listed (wildcards are *NOT* allowed). Expressions are allowed but they *MUST* be labelled.
+| `select isbn, title from books`
+
+| *SELECT SINGLE*
+| The query to use when loading a single entry into the cache. Columns *MUST* be explicitly listed (wildcards are *NOT* allowed). Expressions are allowed but they *MUST* be labelled.
+| `select isbn, title from books where isin = :key`
+
+| *UPSERT* 
+| The query to use when inserting/updating entrys into the database.
+| `insert into books (isbn, title) values() ON DUPLICATE KEY UPDATE`
+
+| *DELETE*
+| The query to use when a single entry is deleted.
+| `delete from books where id = :key`
+
+| *DELETE ALL*
+| The query to use when the cache is cleared.
+| `delete from books/truncate table books`
+
+| *SIZE*
+| The query to use to count the number of entries
+| `select count(*) from books`
+
+|===
+
+The proposed XML configuration for this mode:
+
+[source,xml]
+----
+<local-cache name="books">
+   <persistence>
+      <query-jdbc-store>
+         <data-source jndi-url="..." />
+         <!-- the queries element is required, including the names of all the columns which make up the key -->
+         <queries key-columns="isbn">
+            <select-all>SELECT isbn, title FROM books</select-all>
+            <select-single>SELECT isbn, title FROM books WHERE isbn = :key</select-single>
+            <delete-all>DELETE FROM books</delete-all>
+            <delete>DELETE FROM books WHERE isbn = :key</delete>
+            <upsert>INSERT INTO books (isbn, title) VALUES (:key, :value) ON CONFLICT (isbn) DO UPDATE SET title = :value</upsert>
+            <size>SELECT COUNT(*) FROM books</size>
+         </queries>
+         <!-- the schema element is optional. Defaults will be used otherwise -->
+         <schema message-name="Books" key-message-name="BooksID" file-name="books.proto" package="library"/>
+      </query-jdbc-store>
+   </persistence>
+</local-cache>
+----
+
+==== Query parameters
+
+Query parameters are named using the following conventions:
+
+* simple keys or values use the `:key` or `:value` names
+* composite keys or values use the `:key` or `:value` names followed by a `.` and the name of the field, e.g. `:value.description`
+
+== Storage
+
+=== ProtoBuf schema
+
+==== User-supplied schema
+
+It should be possible for the user to specify her own message type name, which should be present in the schema registry. 
+The names of the fields in the schema **MUST** match the column names of the table.
+
+==== Generated schema
+
+===== Naming
+
+The following naming conventions are used:
+
+* schema file: `*cache-name*-sql-store.proto`
+* package name: `sql-store`
+* key message type name: `*cache-name*_key`
+* value message type name: `*cache-name*_value`
+
+=== Type mapping
+
+==== Simple
+
+Simple type mapping is used when the table/query only returns two columns. The following conventions are used:
+
+* A string will be used for column types which can be represented as such (char, varchar, text, numbers, boolean)
+* a byte array will be used for binary types (blob, bytea)
+
+
+==== ProtoBuf
+
+The following table shows the default mappings between common SQL types and their ProtoBuf equivalents:
+
+[cols="1,1"]
+|===
+| *SQL*
+| *ProtoBuf*
+
+| `int4`
+| `int32`
+
+| `int8`
+| `int64`
+
+| `float4`
+| `float`
+
+| `float8`
+| `double`
+
+| `numeric`
+| `double`
+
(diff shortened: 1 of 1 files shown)


SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+|===
+| *Query*
+| *Description*
+| *Example*
+
+| *SELECT ALL*
+| The query to use when bulk-loading data into the cache. Columns *MUST* be explicitly listed (wildcards are *NOT* allowed). Expressions are allowed but they *MUST* be labelled.
+| `select isbn, title from books`
+
+| *SELECT*
```
danberindei, line 55:
```suggestion
| *SELECT SINGLE*
```

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+[source,xml]
+----
+<local-cache name="default">
+   <persistence>
+      <query-jdbc-store>
+         <data-source jndi-url="..." />
+         <!-- the queries element is required, including the names of all the columns which make up the key -->
+         <queries key-columns="isbn">
+            <select-all>SELECT isbn, title FROM books</select-all>
+            <select>SELECT isbn, title FROM books WHERE isbn = :key</select>
```
danberindei, line 84:
```suggestion
            <select-single>SELECT isbn, title FROM books WHERE isbn = :key</select-single>
```

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+* primitive keys or values use the `:key` or `:value` names
+* composite keys or values use the `:key` or `:value` names followed by a `.` and the name of the field, e.g. `:value.description`
+
+== Storage
+
+=== ProtoBuf schema
+
+==== User-supplied schema
+
+It should be possible for the user to specify her own schema name, which should be present in the schema registry.
```
danberindei, line 109:
```suggestion
It should be possible for the user to specify her own message type name, which should be present in the schema registry. 
```

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+* composite keys or values use the `:key` or `:value` names followed by a `.` and the name of the field, e.g. `:value.description`
+
+== Storage
+
+=== ProtoBuf schema
+
+==== User-supplied schema
+
+It should be possible for the user to specify her own schema name, which should be present in the schema registry. 
+The names of the fields in the schema **MUST** match the column names of the table.
```
danberindei, line 110: Should we allow the fields to be divided into multiple message types, which are then aggregated into a value message type? E.g. if the key has 2 or more fields and the value needs to include the key, the value could aggregate the key message type instead of listing the fields again.

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+==== Generated schema
+
+===== Naming
+
+The default naming of the schema will use the following convention: `*cache-name*-sql-store.proto`.
```
danberindei, line 116: Should include the convention for the package name and the key/value message type names.

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+=== Type mapping
+
+==== Primitive
+
+Primitive type mapping is used when the table/query only returns two columns. The following conventions are used:
+
+* A string will be used for column types which can be represented as such (char, varchar, text, numbers, boolean)
```
danberindei, line 124: I believe "string type mapping" would be a better name. The "primitive" name makes it sound like it would use the primitive types in protobuf (or the primitive wrapper types in Java).

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+);
+----
+
+the default behaviour will cause the `isbn` column to be used as key and the `title` column to be used as value, both as strings. No ProtoBuf schemas will be generated. The following queries will be used (PostgreSQL):
+
+* SELECT: `SELECT isbn, title FROM books WHERE isbn = ?`
+* SELECT-ALL: `SELECT isbn, title FROM books`
+* DELETE: `DELETE FROM books WHERE isbn = ?`
+* DELETE-ALL: `TRUNCATE books`
+* UPSERT: `INSERT INTO books (isbn, title) VALUES (?, ?) ON CONFLICT (isbn) DO UPDATE SET title = ?`
```
danberindei, line 188: I suggest using the `:key` and `:value` parameter names here in the other examples of generated SQL, to match what the user would have to write in the configuration.

SQL Store.adoc
```diff
@@ -0,0 +1,251 @@
...
+[source,xml]
+----
+<local-cache name="default">
+   <persistence>
+      <table-jdbc-store>
+         <data-source jndi-url="..." />
+         <!-- the table element is required -->
+         <table name="books"/>
+         <!-- the schema element is optional. Defaults will be used otherwise -->
+         <schema name="books.proto" key-name="books_id.proto" package="sql_store"/>
```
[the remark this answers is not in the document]
tristantarrant: > Not sure what `key-name` is supposed to be: a reference to another schema file?

It's the schema "name" used for composite keys. Not used for primitive keys.

> I'm not sure if the configuration should reference schema files at all. I would rather the key and value were configured via message type names (including the package name, and possibly allowing primitive types without a MessageWrapper) and the store looked up the message types in all the schemas registered with the cache manager.

Indeed, the intention was for both `name` and `key-name` to be the message type names. I'll amend that. As for primitive types, that is described elsewhere in the document.

SQL Store.adoc
tristantarrant: Done

SQL Store.adoc
tristantarrant: Added the naming conventions

SQL Store.adoc
tristantarrant: Optionally adding the key fields to the value is a good idea. I'll add that possibility.

SQL Store.adoc
```diff
@@ -0,0 +1,256 @@
...
+      <query-jdbc-store>
+         <data-source jndi-url="..." />
+         <!-- the queries element is required, including the names of all the columns which make up the key -->
+         <queries key-columns="isbn">
+            <select-all>SELECT isbn, title FROM books</select-all>
+            <select-single>SELECT isbn, title FROM books WHERE isbn = :key</select-single>
+            <delete-all>DELETE FROM books</delete-all>
+            <delete>DELETE FROM books WHERE isbn = :key</delete>
+            <upsert>INSERT INTO books (isbn, title) VALUES (:key, :value) ON CONFLICT (isbn) DO UPDATE SET title = :value</upsert>
+         </queries>
```
rvansa, line 88: What about a size query?

tristantarrant: Good idea.

tristantarrant: @rvansa added

SQL Store.adoc
tristantarrant: I've gone with the term "simple".

SQL Store.adoc
tristantarrant: I've added a boolean `embedded-key` to the `schema` configuration.

SQL Store.adoc
rvansa: @tristantarrant I also wonder, since the recent updates to the stores have been quite segment-focused, shouldn't there be a possibility to keep the segment ids around?

That's not in sync with the initial story on the top (that sounds like a read-only story but then you expand into a full-fledged configurable sql store) where you start with existing DB - but if the is supposed to be *The JDBC store* it could fit. But then you might start thinking about metadata...

tristantarrant: I shied away from metadata for this iteration, but that would be the next evolutionary step. This would require a separate table with a `FOREIGN KEY` and `ON DELETE CASCADE`.

rvansa: Understood. I guess the segment id comes in a next step, too, as it requires a new set of operations anyway.

SQL Store.adoc
```diff
@@ -0,0 +1,262 @@
+= SQL Store
+
+As a application developer or data grid administrator, I want to preload a cache from a relational database accessed by JDBC - via configuration without writing boilerplate code.
+
+== Architecture
+
+* the store can be either read/write or read-only (in which case it behaves as a loader)
+* the store can work in either single-table mode or in query-mode (via user supplied queries)
+* the store can work with "primitive" keys/values or with complex types defined as protobuf schemas
+* a protobuf schema can either be generated dynamically from the resultset metadata or supplied by the user
```
ryanemerson, line 10: AFAIK protostream currently only allows proto files and marshallers to be generated from Java objects, not the other way round. Is this something we plan to add to ProtoStream? Or will we simply use the google libs here?

SQL Store.adoc
```diff
@@ -0,0 +1,262 @@
...
+[source,xml]
+----
+<local-cache name="books">
+   <persistence>
+      <table-jdbc-store>
+         <data-source jndi-url="..." />
+         <!-- the table element is required -->
+         <table name="books"/>
+         <!-- the schema element is optional. Defaults will be used otherwise -->
+         <schema name="Book" key-name="BookID" package="library"/>
```
ryanemerson, line 35: How do we specify a composite key here? Should it be `key-names="..."`?

SQL Store.adoc
```diff
@@ -0,0 +1,262 @@
...
+The names of the fields in the schema **MUST** match the column names of the table.
+
+==== Generated schema
+
+===== Naming
+
+The following naming conventions are used:
+
+* schema file: `*cache-name*-sql-store.proto`
+* package name: `sql-store`
```
ryanemerson, line 124: We should prepend this with `org.infinispan`

ryanemerson (commented):
I think usages of "Primitive" still need to be updated to ["Simple"](https://github.com/infinispan/infinispan-designs/pull/25/files#r617275354).

SQL Store.adoc
tristantarrant: Generating a protobuf schema requires nothing more than string concatenation. No other libs required.

SQL Store.adoc
tristantarrant: that is just the ProtoBuf message name. See further down for an example.

SQL Store.adoc
tristantarrant: I did think about that, but then it looks as if the schemas (and generated classes) are officially part of Infinispan. I wanted to make it clear that these are in the user's domain.

SQL Store.adoc
ryanemerson: Should it be called "message-name" then?

SQL Store.adoc
ryanemerson: I was thinking of the `application/x-java-object` use-case and forgetting that the loaded key/values are either `String` or `byte[]`.

SQL Store.adoc
ryanemerson: I think that's fair. If they require more they can specify the package explicility in the config.

wfink: From a user perspective there is a use-case where the data should be reloaded at some point of time.
Expiration will not cover this as entries should expire after X hours but that means after load all it expires i.e. 0:00 but the entries are not loaded immediately but on demand which mean they are expiring at different times.
If now entries within the DB are updated K1 might be loaded before and K2 after that and could be inconsistent from an application perspective.
Another aspect is to schedule the refresh at one time, i.e. after the expected DB update and at a time where low throughput is expected.

Would it be a possible feature to have a scheduled refresh which clear and preload the cache (cron like)?

tristantarrant: Something like this should trigger a full near-cache invalidation on clients

tristantarrant: I've created https://issues.redhat.com/browse/ISPN-12999
I've also linked it to https://issues.redhat.com/browse/ISPN-7903 which is an interesting addition

PR #26, opened 2021-05-18
ISPN-12617 Transport Security

tristantarrant:
https://issues.redhat.com/browse/transport_security

### DIFF
--- a/TransportSecurity.asciidoc
+++ b/TransportSecurity.asciidoc
@@ -1,0 +1,81 @@
+== Server Transport Security
+
+JGroups offers various protocols to secure the transport, however they require complex setup which could be simplified by the server security configuration.
+
+== Encryption
+
+=== SSL/TLS
+
+JGroups does not directly offer the use of TLS/SSL for the TCP transport, although it does allow setting a custom `SocketFactory`.
+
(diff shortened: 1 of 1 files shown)


TransportSecurity.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+        <server:encryption>
+            <credential-reference store="credentials" alias="trust"/>
+        </server:encryption>
+    </transport>
+</cache-container>
+----
+
+=== ASYM_ENCRYPT (no)
+
+The complexity of configuring `ASYM_ENCRYPT` doesn't provide any additional value compared to the SSL/TLS approach described above.
```
danberindei, line 65: Maybe it's worth using ASYM_ENCRYPT over TLS to allow UDP/multicast?

TransportSecurity.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+
+By using the server identity of a security realm, we can supply JGroups with an SSL-enabled SocketFactory. The server identity must have both a keystore as well as a truststore so that certificates are mutually trusted.
+
+By using namespaces, we can extend the transport configuration as follows:
+
+[source,xml]
+----
+<cache-container>
+    <transport>
+        <server:encryption>
```
pruivo, line 19: I would rather create an `<encrypt>` element under `<stack>` since this is related to JGroups. And also I could take advantage of it for xsite :)

PR #27, opened 2021-11-17
Dynamic RBAC

tristantarrant:


Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+                    "204": {
+                        "description": "The role was replaced"
+                    },
+                    "404": {
+                        "description": "Role not found"
+                    }
+                }
+            },
+            "delete": {
+                "description": "Remves a role-to-permission mapping",
```
ryanemerson, line 291: "Removes"

ryanemerson (approved):
LGTM.

I really like the design template, especially the OpenAPI definitions for the REST API.

Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+    * @param context
+    */
+   void setContext(AuthorizationMapperContext context);
+
+   /**
+    * Required permissions: <tt>ADMIN</tt>
+    * @param name the name of the role
+    * @return the {@link Role}
+    */
+   CompletionStage<Role> getRole(String name);
```
karesti, line 80: @tristantarrant all non blocking ?

Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+   /**
+    * Required permissions: <tt>ADMIN</tt>
+    * @return all roles handled by this RolePermissionMapper
+    */
+   CompletionStage<Map<String, Role>> getAllRoles();
```
karesti, line 86: What does the key string stand for? I would expect a List here instead of a map

karesti: I guess this is the role name.

Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+[source,java]
+----
+package org.infinispan.security;
+
+public interface MutableRolePermissionMapper extends RolePermissionMapper {
+   /**
+    * Required permissions: <tt>ADMIN</tt>
+    */
+   CompletionStage<Void> addRole(Role role);
```
karesti, line 98: something indicating the role has been added instead of Void ?

Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+   /**
+    * Required permissions: <tt>ADMIN</tt>
+    */
+   CompletionStage<Void> removeRole(String role);
+}
```
karesti, line 104: indicate the role has been removed instead of Void ?

Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+Introduce an `org.infinispan.security.AuthorizationMapperContext` interface that supersedes `org.infinispan.security.PrincipalRoleMapperContext`
+
+== Deprecations
+
+* `org.infinispan.security.PrincipalRoleMapperContext`
+
+== Hot Rod API
+
+N/A
```
karesti, line 115: we don't want to allow role manipulation on hotrod? I'm thinking about the fact that we are able to configure some security or configure caches in startup. Maybe an API for authz manipulation via hotrod is something we might want to consider at a point?

Dynamic RBAC.adoc
```diff
@@ -0,0 +1,320 @@
...
+* `*roles create* --permissions='perm1'[,'perm2'...] _role_` Create a new role
+* `*roles alter* --permissions='perm1'[,'perm2'...] _role_` Alters an existing role
+* `*roles remove* _role_` Remove an existing roleReplace
+* `*roles ls*` List all roles
+
+== Console
+
+The console should provide a "Global roles" view, with the ability to list, add, edit and remove roles.
+
+== Operator
```
karesti, line 318: I would add a way we want to target as well, in the template, to disseminate this feature. Simple tutorial? Blog post? so we add a JIRA afterwards too.

karesti (commented):
some comments

Dynamic RBAC.adoc
tristantarrant: It's the role name and it is required by the `Authorizer` to quickly lookup a role's permissions. Having it as a map all the way is more efficient.

Dynamic RBAC.adoc
tristantarrant: Definitely not. Defining role is a global server admin operation.

PR #31, opened 2022-04-26
Multimap evolutions

karesti:
https://issues.redhat.com/browse/ISPN-13856

### DIFF
--- a/Multimap-As-A-First-Class-Data-Structure.asciidoc
+++ b/Multimap-As-A-First-Class-Data-Structure.asciidoc
@@ -1,0 +1,126 @@
+= Dynamic RBAC
+
+== Tracking Issue
+
+https://issues.redhat.com/browse/ISPN-13853
+
+== Description
+
+Multimap is a data structure that has gotten the attention from Infinispan users.
+The structure today lacks evolutions and support of features and new API of Infinispan:
+
+* REST API support
+* Supporting duplicates
+* Reactive API integration and development
+* Integration with Quarkus through the extension
+* Search and query support
+
+== Compatibility impact
+
+Ensure that the old Remote API and configuration still works.
+
+Under the hood, we will still keep using distributed caches, but we will know with the configuration that those
+caches are meant to behave as multimaps. This way we will be able to change the implementation details in
+the future if the structure becomes even more popular.
+
+Indexing and Query should evolve to detect that the content to be indexed is inside the list/set of the key/value.
+Query will be supported *only* if indexing is enabled.
+
+== Security impact
+
+N/A
+
+== Configuration Schema
+
+[source,xml]
+----
+<xs:complexType name="multimap">
+    <xs:complexContent>
+      <xs:extension base="tns:distributed-cache">
+        <xs:attribute name="duplicates" type="xs:boolean" default="false">
+          <xs:annotation>
+            <xs:documentation>Determines if the multimap supports duplicates in the values. True or false (default).</xs:documentation>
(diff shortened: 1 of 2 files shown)


Multimap-As-A-First-Class-Data-Structure.asciidoc
```diff
@@ -0,0 +1,125 @@
...
+GET /rest/v2/multimaps/
+
+GET /rest/v2/multimaps/{multimapName}?action=???
+
+DELETE /rest/v2/multimaps/{multimapName}/{multimapKey}
+----
+
+* CRUD multimap (providing a configuration to create a multimap, templates not supported)
+* Display Multimap content entries
+* Get values by key
```
tristantarrant, line 92: We also need an equivalent of the `CacheEntry<K, V> Cache.getEntry()` call. Unfortunately this needs to have a different interface per type

```java
// sync
SyncCacheEntryCollection<K, V> MultiMap.getEntry(K key, CacheOptions options);

public interface SyncCacheEntryCollection<K, V> {
   K key();

   CloseableIterable<V> values();

   CacheEntryMetadata metadata();
}
// async
CompletionStage<AsyncCacheEntryCollection<K, V>> MultiMap.getEntry(K key, CacheOptions options);

public interface AsyncCacheEntryCollection<K, V> {
   K key();

   Flow.Publisher<V> values();

   CacheEntryMetadata metadata();
}
// mutiny
Uni<MutinyCacheEntryCollection<K, V>> MultiMap.getEntry(K key, CacheOptions options);

public interface MutinyCacheEntryCollection<K, V> {
   K key();

   Multi<V> values();

   CacheEntryMetadata metadata();
}
```

Multimap-As-A-First-Class-Data-Structure.asciidoc
```diff
@@ -0,0 +1,125 @@
...
+* In Indexing is enabled, multimaps support query
+* CRUD Key/Value. If a key exists we add the value. To remove a key/value we need to send the value as well.
+* CRUD Key/AllValues. We remove the entry completely.
+* Clearing Multimap operation
+* Enable/Disable rebalancing
+* Availability get/set
+* Replace not supported
```
tristantarrant, line 99: I'd mention listener support as well as possible interactions with near-caching

tristantarrant (changes requested):

Multimap-As-A-First-Class-Data-Structure.asciidoc
karesti: Yes, true. There was in the roadmap as well

Multimap-As-A-First-Class-Data-Structure.asciidoc
wburns: This also brings up the question do values each have their own metadata such as expiration? The API as proposed here wouldn't allow that.

karesti: For now not. Do we want that?

karesti: @tristantarrant @wburns how would we include that ?

karesti:
@wburns the expiration the way we implemented multimap, can only be supported by key, not per key value. I believe being able to change it all to support expiration per key/value won't be able to land very soon. I think we might want to either support expiration of keys and all values inside or not support expiration at all for now.

karesti:
@tristantarrant @wburns I merged this one instead of closing it

PR #32, opened 2023-03-08
Cache Selector

tristantarrant:
Design for https://issues.redhat.com/browse/ISPN-14568

### DIFF
--- a/CacheSelector.adoc
+++ b/CacheSelector.adoc
@@ -1,0 +1,346 @@
+= Cache Selector
+
+== Tracking Issue
+
+https://issues.redhat.com/browse/ISPN-14568
+
+== Description
+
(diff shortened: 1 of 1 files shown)


jabolina:
One point that comes to mind. How should we handle overlapping/conflicting selection rules? For example, `key|regex|user.*|cache1` and `key|prefix|use|cache2`.  Not allow such configurations? Have a priority for rules?

tristantarrant:
> One point that comes to mind. How should we handle overlapping/conflicting selection rules? For example, `key|regex|user.*|cache1` and `key|prefix|use|cache2`. Not allow such configurations? Have a priority for rules?

I'm thinking rules should be ordered

tristantarrant:
I've had a change of heart regarding where this should be configured: this should be part of core and not part of the connectors. I'll revise the proposal

PR #33, opened 2023-04-13
Hash Data Structure

jabolina:
Adding the initial design draft for the Hash data structure. I think there are some unknown unknowns for me here, but I tried to cover at least a bit of everything.

### DIFF
--- a/Cache-Hash-Data-Structure.adoc
+++ b/Cache-Hash-Data-Structure.adoc
@@ -1,0 +1,209 @@
+= Hash Data Structure
+
+Provide a Hash data structure to Infinispan users. The Hash is a flexible identifiable grouping of keys mapping to values.
+In this context, _flexible_ means it can have arbitrary mappings without needing a schema beforehand; _identifiable group_
+means the Hash structure receives a unique name and retrieves the complete key-value mapping.
+
+The Hash provides a general approach for users to represent their data structures using Infinispan. For example, by
+employing this, users can avoid using key prefixes to create an object in Cache.
+
+== Public API
+
+The work is for the new API only. The Hash data structure has the format of:
+
+[source,java]
+.Hash.java
+----
+public interface Hash<K, V> { // <1>
+
+   String getName();
+
+   V get(K key);
+
+   V put(K key, V value);
+
+   V remove(K key);
+
+   boolean containsKey(K key);
+
+   Set<K> keySet();
+
+   Collection<V> values();
+
+   Set<Map.Entry<K, V>> entrySet();
+
+   void clear();
+
+   int size();
+}
+----
+
+<1> The Hash data structure is a generic interface over `K` and `V`. We do not extend `Map<K, V>`, instead we add methods
+as needed.
+
+A `Hash<?, ?>` structure is local to a node. Meaning that operations applied to the local copy must be submitted
+through the public API before taking effect. After that, the `Hash<?, ?>` is distributed/replicated according to
+the configuration.
+
+We follow a state-based CRDT approach, merging the local Hash with the server's Hash. Since the `Hash<?, ?>` structures
+occupy a single entry in the Cache, the whole structure has a version instead of versioning per Hash property. Taking
+advantage of this, we can use the `compute*` method and directly apply the local changes. If a concurrent change happens,
+the `Hash<?, ?>` version is updated, and the replace fails, where the `compute*` retries again until it succeeds. Note
+this could lead to flooding the network, transmitting the `Hash<?, ?>` back and forth, but this only happens with heavy
+concurrent changes over the same structure.
+
+[IMPORTANT]
+.Inconsistency!
+====
+The use of `Hash<?, ?>` could return an outdated value, since the structure is a local copy.
+====
+
+To interact with the `Hash<K, V>`:
+
+[source, java]
+.HashCache.java
+----
+public interface HashCache<K> {
+
+   <HK, HV> HV put(K key, HK hashKey, HV hashValue);
+
+   <HK, HV> Hash<HK, HV> put(K key, Map<HK, HV> values);
+
+   <HK, HV> Map<HK, HV> get(K key, HK hashKey, HK ... others); // <1>
+
+   Hash<HK, HV> get(K key);
+
+   boolean remove(K key);
+
+   <HK> boolean remove(K key, HK hashKey, HK ... others);
+
+   boolean exists(K key);
+
+   <HK> boolean containsKey(K key, HK hashKey);
+
+   <HK, HV> Hash<HK, HV> submit(Hash<HK, HV> hash); // <2>
+}
+----
+
+<1> Returning a `Map<HK, HV>` to contain all the vargs keys.
+<2> The `submit` method is used to apply the local changes to the server's Hash. The server's `Hash<?, ?>` is returned.
+
+I see two options for the operation applying directly to a `Hash<?, ?>`, like `put` and `get`:
+
+1. The interface is just a convenient method. The concrete implementation internally calls the `submit` method,
+effectively transferring the `Hash<?, ?>` to the client, applying the operation, and submitting it to the server.
+2. We implement custom commands for these operations and rely on `compute*` only for the `submit` method. Although,
+unsure if this is feasible because the core would need to know about the `Hash<?, ?>` structure.
+
+And as of Multimap, we provide APIs for:
+
+=== Hot Rod API
+
+Developed only for the new API. Provide implementations for an `AsyncHashCache` and `MutinyHashCache`.
+
+=== REST API
+
+Hash REST API should be the almost the same as the Cache one.
+The difference is in the endpoint.
+
+Examples:
+[source]
+----
+GET /rest/v2/hashes/
+
+GET /rest/v2/hashes/{hashName}?action=???
+
+DELETE /rest/v2/hashes/{hashName}/{hashKey}
+
+DELETE /rest/v2/hashes/{hashName}/{hashKey}/{hashProperty}
+----
+
+* CRUD hashes (providing a configuration to create a hash, templates not supported)
+* Display Hash content entries
+* Get values by key
+* In Indexing is enabled, hashes support query
+* CRUD for operations available in the `HashCache<?>` API
+* Enable/Disable rebalancing
+* Availability get/set
+* Replace not supported
+* Listeners support to enable interactions with near-caching
+
+
+== Using the Hash
+
+A new maven module is created.
+
+=== Configuration Schema
+
+[source,xml]
+----
+<xs:complexType name="hashes">
+    <xs:sequence>
+        <xs:element name="hash" type="hash" minOccurs="0" maxOccurs="unbounded">
+            <xs:annotation>
+                <xs:documentation>
+                    Defines a hash data structure.
+                </xs:documentation>
+            </xs:annotation>
+        </xs:element>
+    </xs:sequence>
+</xs:complexType>
+
+<xs:complexType name="hash">
+    <xs:attribute name="name" type="xs:ID" use="required">
+        <xs:annotation>
+            <xs:documentation>
+                A unique name for the hash.
+            </xs:documentation>
+        </xs:annotation>
+    </xs:attribute>
+</xs:complexType>
+----
+
+To create and use a `HashCache<?>`:
+
+```java
+try (Infinispan ispn = Infinispan.create(builder.build())) {
+   Hashes hashes = ispn.sync().hashes(); // <1>
+   HashCache<String> users = hashes.get("users");
+
+   users.put("42", Map.of("auth", "token", "role", "user"));
+   System.out.println(hashCache.get("42")); // {"auth": "token", "role": "user"}
+}
+```
+
+<1> We provide an implementation for `sync()`, `async()`, and `mutiny()`.
+
+Alternatively, applying multiple operation to the `Hash<?, ?>`, should be:
+
+```java
+try (Infinispan ispn = Infinispan.create(builder.build())) {
+   Hashes hashes = ispn.sync().hashes();
+   HashCache<String> users = hashes.get("users");
+
+   Hash<String, String> hash = users.get("42");
+
+   hash.put("auth", "token");
+   hash.put("role", "user");
+
+   users.submit(hash);
+
+   System.out.println(hashCache.get("42")); // {"auth": "token", "role": "user"}
+}
+```
+
+After applying the operations, then submits the `Hash<String, String>` to the server. Note that in an environment with
+concurrent updates applied to the "users" Hash with key "42", the values could be different already when it is retrieved.
+
+== Infinispan Server Hash
+
+We can build everything on top of a `Cache<K, Hash<HK, HV>>`, using the existing operations available.
+A few open questions are:
+
+* (How) Do we provide the complete set of Cache configurations to Hash? Since we build on top of Cache, it seems
+  reasonable to have the whole configuration options available.
+* Does the expiration apply to a key or a property within the Hash? It seems easier to have the expiration set
+  for the whole Hash. At this point, we're going with an expiration for the whole Hash.
+* Can we integrate this with querying and indexing? I understand we do not provide this over dynamic objects.
+
+
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This repository contains various design proposals related to Infinispan
 * [Index affinity proposal](Index-affinity-proposal.asciidoc)
 * [CLI](Infinispan-CLI.asciidoc)
 * [Hibernate Second-Level Cache improvements](Infinispan-Hibernate-Second-Level-Cache-improvements.md)
+* [Hash Cache Data Structure](Cache-Hash-Data-Structure.adoc)
 * [Query - Design and Planning](Infinispan-Query---Design-and-Planning.asciidoc)
 * [Query language syntax and considerations](Infinispan-query-language-syntax-and-considerations.asciidoc)
 * [Java 8 API proposal](Java-8-API-proposal.md)


Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,174 @@
...
+employing this, users can avoid using key prefixes to create an object in Cache.
+
+== Public API
+
+The work is for the new API only. The Hash data structure has the format of:
+
+[source,java]
+.Hash.java
+----
+public interface Hash<K, V> { // <1>
```
karesti, line 17: How is this different from cache ?

tristantarrant: Because it is stored as a single entry in a cache.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,174 @@
...
+   Set<Map.Entry<K, V>> entrySet();
+
+   void clear();
+
+   int size();
+}
+----
+
+<1> The Hash data structure is a generic interface over `K` and `V`. Should we extend `Map<K, V>`?
+This way we could provide a wide range of operations over the Hash.
```
tristantarrant: I would avoid extending `Map` as that has a million methods that have made it a pain in the past for our `Cache`.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,174 @@
...
+            </xs:documentation>
+        </xs:annotation>
+    </xs:attribute>
+</xs:complexType>
+----
+
+To create and use a `HashCache<?>`:
+
+```java
+      EmbeddedCacheManager cm = ...
```
karesti: new api ?

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,174 @@
...
+      HashCacheManager hashCacheManager = EmbeddedHashCacheManagerFactory.from(cm);
```
karesti: this is the legacy API, we need this in the new API only

Cache-Hash-Data-Structure.adoc
karesti: yes, why not using map directly if we do that.

Cache-Hash-Data-Structure.adoc
karesti: ok, now I get it

karesti (changes requested):

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,174 @@
...
+   int size();
+}
+----
+
+<1> The Hash data structure is a generic interface over `K` and `V`. Should we extend `Map<K, V>`?
+This way we could provide a wide range of operations over the Hash.
+
+A `Hash<?, ?>` structure is local to a node. Meaning that operations applied to the local copy must be submitted
+through the public API before taking effect. After that, the `Hash<?, ?>` is distributed/replicated according to
+the configuration.
```
rigazilla, line 46: How is the intended use case?
The user operates on an Hash object and then uses an HashCache obj to submit the changes?

tristantarrant: I would say that individual operations are applied immediately to the underlying storage (remote cache). When using tx/batching, these would be collected and sent in a single operation.

Cache-Hash-Data-Structure.adoc
tristantarrant: Some further information:
* `Map` is a sync-only API, so this does not help with the async and mutiny variants
* Some of the `Map` methods have effects, such as returning the previous value on `put` and `remove` which are costly in remote scenarios.

Cache-Hash-Data-Structure.adoc
tristantarrant: Let's not implement anything on the legacy API please.

Cache-Hash-Data-Structure.adoc
jabolina: > The user operates on an Hash object and then uses an HashCache obj to submit the changes?

Yeah, I was thinking of something of that sort. I'll update the documentation to include single operations without transferring the whole object. For multiple operations, I was thinking of keeping track of the local changes, and once the user submits the changes to the server, we merge the local Hash with the server's Hash.

I'll update the docs.

jabolina:
Thanks, everyone, for the comments. I updated the document with all the suggestions. While at it, I described more about how we handle the operations. A point of attention is how we should handle operations that directly reference a Hash, for example, `HashCache.put("hash", "property", "value")`, I added the two options I see for that in the document.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+   boolean exists(K key);
+
+   <HK> boolean containsKey(K key, HK hashKey);
+
+   <HK, HV> Hash<HK, HV> submit(Hash<HK, HV> hash); // <2>
+}
+----
+
+<1> Returning a `Map<HK, HV>` to contain all the vargs keys.
+<2> The `submit` method is used to apply the local changes to the server's Hash. The server's `Hash<?, ?>` is returned.
```
rigazilla, line 89: Do we really need the return value? Hash could be very big. Returning on request or only if submit fails?

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+   void clear();
+
+   int size();
+}
+----
+
+<1> The Hash data structure is a generic interface over `K` and `V`. We do not extend `Map<K, V>`, instead we add methods
+as needed.
+
+A `Hash<?, ?>` structure is local to a node. Meaning that operations applied to the local copy must be submitted
```
rigazilla, line 44: I feel a disturbance about Hash. I mean, if it's detached from cache isn't it just a Map do we need a new type? On the other side if individual operations are submitted immediately then some methods seems to overlap with HashCache

wburns: I agree, I did not like the naming on the RESP commands either. A Hash to me is not a Map structure but rather a numeric derived from some other fields.

Cache-Hash-Data-Structure.adoc
wburns: Looking at this, it looks awfully close our MultiMap, just instead of a Collection it is a Map. Could we not build on top of that?

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+====
+
+To interact with the `Hash<K, V>`:
+
+[source, java]
+.HashCache.java
+----
+public interface HashCache<K> {
+
+   <HK, HV> HV put(K key, HK hashKey, HV hashValue);
```
wburns, line 68: We should probably name this method `getAndPut` and have just a `put` method that has no returned.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+====
+The use of `Hash<?, ?>` could return an outdated value, since the structure is a local copy.
+====
+
+To interact with the `Hash<K, V>`:
+
+[source, java]
+.HashCache.java
+----
+public interface HashCache<K> {
```
wburns, line 66: ~~This entire interface is blocking, we probably need NonBlocking versions first.~~

Reading further it looks like we would have all 3 of the variants 👍

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+   <HK, HV> HV put(K key, HK hashKey, HV hashValue);
+
+   <HK, HV> Hash<HK, HV> put(K key, Map<HK, HV> values);
```
wburns, line 70: Same here, we really need just a `put` and `getAndPut`.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+   boolean remove(K key);
+
+   <HK> boolean remove(K key, HK hashKey, HK ... others);
+
+   boolean exists(K key);
+
+   <HK> boolean containsKey(K key, HK hashKey);
+
+   <HK, HV> Hash<HK, HV> submit(Hash<HK, HV> hash); // <2>
```
wburns, line 84: Confused as to what this method actually does. Why is there no key tied to the hash?

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+   <HK, HV> Hash<HK, HV> put(K key, Map<HK, HV> values);
+
+   <HK, HV> Map<HK, HV> get(K key, HK hashKey, HK ... others); // <1>
+
+   Hash<HK, HV> get(K key);
+
+   boolean remove(K key);
+
+   <HK> boolean remove(K key, HK hashKey, HK ... others);
```
wburns, line 78: A single boolean doesn't work here as some may be removed and some may not.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+DELETE /rest/v2/hashes/{hashName}/{hashKey}
+
+DELETE /rest/v2/hashes/{hashName}/{hashKey}/{hashProperty}
+----
+
+* CRUD hashes (providing a configuration to create a hash, templates not supported)
+* Display Hash content entries
+* Get values by key
+* In Indexing is enabled, hashes support query
```
wburns, line 124: How would it do this? Are we planning on having the hashes have their own key and value schema definitions? I don't see that mentioned.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+* CRUD for operations available in the `HashCache<?>` API
+* Enable/Disable rebalancing
+* Availability get/set
+* Replace not supported
+* Listeners support to enable interactions with near-caching
```
wburns, line 129: Listeners are going to be very very very expensive for this without some sort of delta notification, which we do not have for listeners. Also listener API was never solidified for the new API sadly :(

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+To create and use a `HashCache<?>`:
+
+```java
+try (Infinispan ispn = Infinispan.create(builder.build())) {
+   Hashes hashes = ispn.sync().hashes(); // <1>
+   HashCache<String> users = hashes.get("users");
+
+   users.put("42", Map.of("auth", "token", "role", "user"));
+   System.out.println(hashCache.get("42")); // {"auth": "token", "role": "user"}
```
wburns, line 171: Is `hashCache` here supposed to be `users`?

Also from this code, I assume that an updates to the returned `HashCache` instance are immediately updated remotely and applied to the `Hashes` actual contents? We will need to make sure the interface is documented that that is done that way.

Only the `Hash` interface is the one that isn't propagated correct?

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+<1> We provide an implementation for `sync()`, `async()`, and `mutiny()`.
+
+Alternatively, applying multiple operation to the `Hash<?, ?>`, should be:
+
+```java
+try (Infinispan ispn = Infinispan.create(builder.build())) {
+   Hashes hashes = ispn.sync().hashes();
+   HashCache<String> users = hashes.get("users");
+
+   Hash<String, String> hash = users.get("42");
```
wburns, line 184: So the `Hash` returned will always be not null I assume?

wburns: This makes me wonder if we should have some sort of putHashIfAbsent type of operation, where you can do an initial request to populate the Hash without having to do an initial get round trip.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+   int size();
+}
+----
+
+<1> The Hash data structure is a generic interface over `K` and `V`. We do not extend `Map<K, V>`, instead we add methods
+as needed.
+
+A `Hash<?, ?>` structure is local to a node. Meaning that operations applied to the local copy must be submitted
+through the public API before taking effect. After that, the `Hash<?, ?>` is distributed/replicated according to
```
wburns, line 45: I am debating about this. It seems we could also implement this in a slightly different fashion where the key is just a name of a cache and the underlying Hash is just a normal Cache which itself can be distributed etc. This would then give us delta changes between keys and we can use all of the existing features of Caches. I am not quite sold yet, especially as I have no idea what the query API would look like or listeners.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+try (Infinispan ispn = Infinispan.create(builder.build())) {
+   Hashes hashes = ispn.sync().hashes();
+   HashCache<String> users = hashes.get("users");
+
+   Hash<String, String> hash = users.get("42");
+
+   hash.put("auth", "token");
+   hash.put("role", "user");
+
+   users.submit(hash);
```
wburns, line 189: So will this completely overwrite the hash if it is already exists or only apply the new values? It isn't clear or doesn't seem defined yet as to what this would do. I am guessing it completely overwrites those values, in which case you would lose any concurrent writes without a respective remove.

Cache-Hash-Data-Structure.adoc
jabolina: We could build upon MultiMap without duplicates, storing key-value pairs. But I am unsure if we can compare for equality based only on the key. On the other hand, maybe we could abstract some of the MultiMap implementations which work over a Collection to work on another structure instead of creating everything from scratch. I would need to look at that, though.

Cache-Hash-Data-Structure.adoc
jabolina: To be completely honest, even after writing the word "Hash" so many times, it still sounds weird as a structure name. I thought the same, i.e., a hash is a value after applying a hashing function (or even the function)... Regarding naming, I would not mind changing it to Map(ping), Document, Structure, Container, or something else.

And as the structure Hash, which the user interacts with, would help to keep track of the operations. For example, if we return a Map and the user removes one of the keys, we wouldn't notice that when merging with the server's content. Or, to better reflect, rename the structure HashView, HashCopy, or something of that sort.

Cache-Hash-Data-Structure.adoc
jabolina: Yeah, but not strictly. I'll add some void methods, too. In one case, for example, the RESP connector needs to return what has changed. But I was aiming at something more general instead of focusing only on the RESP needs.

Cache-Hash-Data-Structure.adoc
jabolina: This one is straight up a stretch, and one thing I had doubts about (the last bullet point at the end of the doc). If feasible, we would have something similar to the [JSON support](https://redis.io/docs/stack/json/). Should I mark it optional? Or remove?

Cache-Hash-Data-Structure.adoc
jabolina: I'll mark it optional. When we establish the listener API, we can come back to this.

Cache-Hash-Data-Structure.adoc
jabolina: > Is `hashCache` here supposed to be `users`?

Yes, I'll fix it.


> Also from this code, I assume that an updates to the returned `HashCache` instance are immediately updated remotely and applied to the `Hashes` actual contents? We will need to make sure the interface is documented that that is done that way.
> 
> Only the `Hash` interface is the one that isn't propagated correct?

Correct. Only the operations directly over the `HashCache` takes place immediately. If operating over a `Hash` it needs to be sent to the `HashCache` to take effect.

I've been thinking more about this and I'll do an update to the `HashCache` API.

Cache-Hash-Data-Structure.adoc
jabolina: For this one, I was thinking to implement with compute and merge the values. We keep track of the operations with the `Hash`, but effectively all the local changes take precedence. I'll do an update in the docs and API.

jabolina:
@rigazilla and @wburns, thanks for the comments! I am going to update the doc and the `HashCache` API introducing your suggestions. I'll try to document better the intents with the API, evidencing the `Hash`needs to be sent to take effect. And rename that `submit` method to something meaningful.

Cache-Hash-Data-Structure.adoc
pruivo: isn't it the same as `FineGrainedAtomicMap` that we dropped and replaced with the `Grouping API`?

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+== Public API
+
+The work is for the new API only. The Hash data structure has the format of:
+
+[source,java]
+.Hash.java
+----
+public interface Hash<K, V> { // <1>
+
+   String getName();
```
pruivo, line 19: what's the name? don't see any reference to it.

Cache-Hash-Data-Structure.adoc
```diff
@@ -0,0 +1,209 @@
...
+----
+
+<1> The Hash data structure is a generic interface over `K` and `V`. We do not extend `Map<K, V>`, instead we add methods
+as needed.
+
+A `Hash<?, ?>` structure is local to a node. Meaning that operations applied to the local copy must be submitted
+through the public API before taking effect. After that, the `Hash<?, ?>` is distributed/replicated according to
+the configuration.
+
+We follow a state-based CRDT approach, merging the local Hash with the server's Hash. Since the `Hash<?, ?>` structures
```
pruivo, line 48: how are we going to merge when network partitions happen? Are we going to support cross-site?

PR #34, opened 2024-06-25
Adding proposal for deadlock detection and resolution

jabolina:
A proposal detailing an algorithm for deadlock detection. We would implement it without changing the behavior of the transactions or the locking mechanism. The resolution is rolling back the transaction if there is a deadlock.

The proposal goes on a high level to implement it for ISPN. Hopefully, most of the mechanisms necessary are in place.


jabolina:
The biggest caveat is the time for probing. Even if it identifies the deadlock and rollback one of the transactions, the operation already took a long time (15s, for example). Should we aim for something that could generate the graph during the lock acquisition phase?

pruivo:
What benefit does it have over configuring a shorter lock acquisition timeout? 

We had a deadlock detection in place but only detected simple and direct graphs between two transactions. If you have a larger cycle, like `tx1 -> tx2 -> tx3 -> tx1`, it wouldn't work. 

So I have to ask. What is the cost/benefit of this approach? Would be better to use a random timeout value that would be assigned per request?

jabolina:
I believe the win with the algorithm instead of random timeouts would be to avoid aborting transactions unnecessarily. However, it is much more intricate to implement than random timeouts in an interval.

I'll research to see if we can adapt the current proposal to run during the lock acquisition to build the graph to identify the cycles. This way, we would proactively deal with deadlocks instead of waiting for timeouts. The responsiveness would be much better than random timeouts or probing after timeouts.

jabolina:
@pruivo, I've updated the proposal. The idea is to have the deadlock detection running concurrently with the transactions. The transaction does not depend on the deadlock to finish, so the additional overhead is just because of more messages in the system. The algorithm should be able to identify the cycles and roll back only the needed transactions to ensure progress.

Because of the figures and graphs, I think it is easier to read it on my branch (https://github.com/infinispan/infinispan-designs/blob/9ce4efed9de250df7432b8eac2f39a1a72b3f2b7/deadlock-detection-resolution.md) instead of the diff and add the comments here.

jabolina:
Just so I don't forget. I need to add details about partition handling and topology changes. :writing_hand:

PR #35, opened 2025-02-05
Schemas api design

karesti:
Schemas API for 16.0

### DIFF
--- a/Schemas-API.adoc
+++ b/Schemas-API.adoc
@@ -36,3 +36,29 @@ protostreamMarshaller.register(schema, magazineMarshaller, bookMarshaller);
 
 builder.remoteCache("cacheName").marshaller(marshaller);
 ----
+
+### Administration API
+
+[source, java]
+----
+public interface RemoteCacheManagerAdmin {
+    SchemasAdministration schemas();
+}
+----
+
+[source, java]
+----
+public interface SchemasAdministration {
+    CompletionStage<SchemaOpResult> createAsync(Schema schema);
+    SchemaOpResult create(Schema schema);
+
+    CompletionStage<SchemaOpResult> updateAsync(Schema schema);
+    SchemaOpResult update(Schema schema);
+
+    CompletionStage<SchemaOpResult> deleteAsync(String name);
+    SchemaOpResult delete(String name);
+
+    CompletionStage<SchemaOpResult> validateAsync(Schema schema);
+    SchemaOpResult validate(Schema schema);
+}
+----
--- a/Schemas-API.adoc
+++ b/Schemas-API.adoc
@@ -134,10 +134,16 @@ public interface SchemasAdministration {
     CompletionStage<SchemaOpResult> deleteAsync(String name);
     SchemaOpResult delete(String name);
 
-    CompletionStage<SchemaOpResult> validateAsync(Schema schema);
-    SchemaOpResult validate(Schema schema);
-
     CompletionStage<SchemaErrors> retrieveSchemaErrors();
+
     SchemaErrors retrieveSchemaErrors(Schema schema);
+    // # Optional. Check if this is needed or possible. Not creating a schema and validating it only.
+        CompletionStage<SchemaOpResult> validateAsync(Schema schema);
+        SchemaOpResult validate(Schema schema);
+    // # end of optional
 }
 ----
+
+SchemaOpResult: Specify the content and return values when implementing it
+based on real code examples around (like keycloak or what quarkus or spring users do)
+
--- a/Schemas-API.adoc
+++ b/Schemas-API.adoc
@@ -37,6 +37,79 @@ protostreamMarshaller.register(schema, magazineMarshaller, bookMarshaller);
 builder.remoteCache("cacheName").marshaller(marshaller);
 ----
 
+### Administration api today
+Hotrod specific api does not exist. However, an existing REST API exists.
+
+Example of what clients are doing today to manage errors:
+
+[source, java]
+----
+    public static void uploadAndReindexCaches(RemoteCacheManager remoteCacheManager, GeneratedSchema schema, List<IndexedEntity> indexedEntities) {
+        var key = schema.getProtoFileName();
+        var current = schema.getProtoFile();
+
+        var protostreamMetadataCache = remoteCacheManager.<String, String>getCache(InternalCacheNames.PROTOBUF_METADATA_CACHE_NAME);
+        var stored = protostreamMetadataCache.getWithMetadata(key);
+        if (stored == null) {
+            if (protostreamMetadataCache.putIfAbsent(key, current) == null) {
+                logger.info("Infinispan ProtoStream schema uploaded for the first time.");
+            } else {
+                logger.info("Failed to update Infinispan ProtoStream schema. Assumed it was updated by other Keycloak server.");
+            }
+            checkForProtoSchemaErrors(protostreamMetadataCache);
+            return;
+        }
+        if (Objects.equals(stored.getValue(), current)) {
+            logger.info("Infinispan ProtoStream schema is up to date!");
+            return;
+        }
+        if (protostreamMetadataCache.replaceWithVersion(key, current, stored.getVersion())) {
+            logger.info("Infinispan ProtoStream schema successful updated.");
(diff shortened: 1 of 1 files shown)


Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+### Client side api proposal
+
+[source, java]
+----
+Schema schema = // programmatic schema
+MagazineMarshaller magazineMarshaller = new MagazineMarshaller(); // custom marshaller 1
+BookMarshaller bookMarshaller = new BookMarshaller(); // custom marshaller 2
+
+ProtoStreamMarshaller protostreamMarshaller = new ProtoStreamMarshaller();
+protostreamMarshaller.register(schema, magazineMarshaller, bookMarshaller);
```
tristantarrant, line 35: Make the `ProtoStreamMarshaller` methods `return this` so they can be chained:

`ProtoStreamMarshaller protostreamMarshaller = new ProtoStreamMarshaller().register(schema, magazineMarshaller, bookMarshaller);`

Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+builder.remoteCache("cacheName").marshaller(marshaller);
```
tristantarrant, line 37: When retrieving a cache with a `ProtoStreamMarshaller` for the first time, would you also invoke `SchemasAdministration.createAsync()` ?

Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+----
+public interface RemoteCacheManagerAdmin {
+    SchemasAdministration schemas();
+}
+----
+
+[source, java]
+----
+public interface SchemasAdministration {
+    CompletionStage<SchemaOpResult> createAsync(Schema schema);
```
tristantarrant, line 125: I think I'd prefer a single `createOrUpdate()` method. The server-side validation will ensure that the new schema is compatible with the old one using the protolock rules

tristantarrant (changes requested):
Some suggestions

Schemas-API.adoc
karesti: I guess that is the implementation, yes, but only if we enable it

karesti: I need to look at the code for that, to make it possible

Schemas-API.adoc
karesti: but if we need to be created only if it does not exist ?

Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+    SchemaOpResult create(Schema schema);
```
jabolina, line 126: Sorry, maybe it exists already, but what does the `SchemaOpResult` hold? Would it help identify a schema violation or something similar?

Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+serializationContext.registerProtoFiles(fds);
+serializationContext.registerMarshaller(magazineMarshaller);
+serializationContext.registerMarshaller(bookMarshaller);
+builder.marshaller(protostreamMarshaller);
+builder.remoteCache("cacheName").marshaller(protostreamMarshaller);
+----
+
+### Client side api proposal
+
+[source, java]
```
jabolina, line 28: This example doesn't seem to be utilizing the `SchemasAdministration` API, is that correct?

Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+----
+public interface SchemasAdministration {
+    CompletionStage<SchemaOpResult> createAsync(Schema schema);
+    SchemaOpResult create(Schema schema);
+
+    CompletionStage<SchemaOpResult> updateAsync(Schema schema);
+    SchemaOpResult update(Schema schema);
+
+    CompletionStage<SchemaOpResult> deleteAsync(String name);
+    SchemaOpResult delete(String name);
```
jabolina, line 135: Hmmm, delete seems quite dangerous. If the schema is removed, the data in a cache would only be accessible as `byte[]` afterward. What would be the use cases for removing a schema? I understand it exists in the current API as well, but maybe we could hide some dangerous operations from users.

jabolina (commented):
Just a few questions, but it looks good to me. I guess we then eventually make it impossible for the user to retrieve and operate the schemas cache?

Schemas-API.adoc
pruivo: It would be helpful to have a `createOrUpdate()` method for Keycloak. Additionally, it would be useful to have a way to detect changes in the index model as well. 

Keycloak is doing the following at the moment, and I would love to simplify things :smile: 
https://github.com/keycloak/keycloak/blob/636fffe0bc37a63bd5a2578b6bbcd815364c41d8/model/infinispan/src/main/java/org/keycloak/marshalling/KeycloakIndexSchemaUtil.java#L71-L97

To be honest, I would love to get rid of `KeycloakIndexSchemaUtil` from Keycloak.

Schemas-API.adoc
wburns: It would technically make any cache that has their encoding set to protostream no longer accessible even as it couldn't deserialize the values. You would also have to update any cache to be octet-stream or unknown encoding iirc.

Schemas-API.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+    CompletionStage<SchemaOpResult> validateAsync(Schema schema);
```
wburns: Is this not validated on the create?

Schemas-API.adoc
jabolina: Yeah, I wasn't sure if it would make it completely inaccessible. Would it be enough if the client changes the request type, or does it need to update the cache configuration? I don't know if encoding can be updated at runtime. But either way, after that, all the data would become just bytes, and the client application would be responsible for handling the encoding.

Schemas-API.adoc
ryanemerson: I think this is a good idea if we have `SchemasAdministration#createOrUpdate`. It should also be possible for clients to opt-out of this behaviour. For example, what happens if a client doesn't have sufficient authorization privileges to perform the action?

Schemas-API.adoc
karesti: yes, this is a great suggestion / request

Schemas-API.adoc
karesti: no, this is for client side only. schemas api is for client/server and it does not exist at all, since we use directly cache api of the PROTOBUF_METADATA_CACHE. I'll add it so it's cleared

Schemas-API.adoc
karesti: I don't know exactly yet, before actually implementing it

karesti: I need to really know checking the actual use real people do and need these days in real code

Schemas-API.adoc
karesti: We are already exposing this through every other api (rest api, actual hotrod api). I don't think we need to hide this from users, that would just make people still use the delete api available in the cache api and make their life harder. We need to document this well though.

karesti:
> Member

that would break a lot of code, I don't think we should do that to be honest

Schemas-API.adoc
karesti: it is, it is a method that eventually would validate a schema without creating it. I'm not sure we need it and it is possible that we can't do it properly, so it's more like an optional api. I'll comment that in the design

karesti:
I added some more info, things might change on the real api implementation, but that can be reviewed for real in the PR

jabolina (approved):

rigazilla (approved):

wburns (approved):

Schemas-API.adoc
```diff
@@ -0,0 +1,149 @@
...
+[source, java]
+----
+public interface SchemasAdministration {
+    CompletionStage<SchemaOpResult> createAsync(Schema schema);
+    SchemaOpResult create(Schema schema);
+
+    CompletionStage<SchemaOpResult> updateAsync(Schema schema);
+    SchemaOpResult update(Schema schema);
+
+    CompletionStage<SchemaOpResult> createOrUpdate(Schema schema);
```
rigazilla:
```suggestion
    CompletionStage<SchemaOpResult> createOrUpdateAsync(Schema schema);
```

PR #36, opened 2026-03-25
Encoding and Rolling Upgrade improvements

wburns:
Design document for providing additional encoding/marshaller updates and also to allow for Rolling Upgrades where users can change marshaller/media type. This will ease users upgrades in both embedded and server allowing for changing of media type and also better controls as to what is stored in the cache.

### DIFF
--- a/Marshalling-Upgrades.adoc
+++ b/Marshalling-Upgrades.adoc
@@ -1,0 +1,65 @@
+= System Upgrades & Marshalling Architecture
+
+== Marshalling Architecture & Exceptions
+
+=== Rest Cache Exception (v1.0)
+
+Implement a hardcoded exception for the "Rest Cache" to bypass broken media type encoding. This prevents standard validation from failing a known-broken component during the transition.
+
+=== Named Marshallers
+
+Implement named marshallers that can be overwritten at the media type level.
+
+=== Conflict Check
+
+If Storage and Request media types both have defined marshallers and they do not match, the system must throw an error.
+
+== Cluster Heterogeneity & Join Logic
+
+=== Relaxed Join Validation
+
+If a node joins with a different media type than the existing cluster, allow it to proceed without error. This ensures the cluster can stabilize during the 1.0 transition phase.
+
+=== Media-Type Aware Commands
+
+All internal commands must account for the specific storage media types of target nodes and perform dynamic conversion.
+
+=== Optimization
+
+Elide (skip sending) the media type metadata if the cluster is confirmed to be homogeneous.
+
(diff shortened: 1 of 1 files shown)


Marshalling-Upgrades.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+== Cluster Heterogeneity & Join Logic
+
+=== Relaxed Join Validation
+
+If a node joins with a different media type than the existing cluster, allow it to proceed without error. This ensures the cluster can stabilize during the 1.0 transition phase.
+
+=== Media-Type Aware Commands
+
+All internal commands must account for the specific storage media types of target nodes and perform dynamic conversion.
```
jabolina, line 25: Are you planning to transfer the information in the commands or handle it in the interceptor? 

The nodes could exchange the information once they join, and it could be handled in the interceptor chain, or maybe in the `EncoderCache`. An approach with an interceptor could be self-contained; the interceptor layer would perform the conversion, so the other parts could be oblivious when requesting/replicating remotely.

Marshalling-Upgrades.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+If a Storage Layer is defined, default to bit-for-bit persistence to avoid transcoding overhead.
+
+=== Rule 2 (Fallback)
+
+If no storage layer is defined, use the default user-defined marshaller.
+
+=== Rule 3 (The Override Warning)
+
+Log a performance warning if the user overrides both the persistence and storage layers with different configurations, as this forces constant transcoding.
```
jabolina, line 43: This seems worth adding to main even before adding to the feature branch.

Marshalling-Upgrades.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== Key Equality & Identity
+
+Use a Key Wrapper or Identity Resolver to bridge the gap between binary keys (Persistence) and Java Object keys (Storage) to ensure equality checks don't fail.
```
jabolina, line 47: This part is interesting. I guess we should have something in place to ensure the key media type is the same globally? We allow only value in this mix.

jabolina (commented):
I am also curious about some operations that require iteration (and state transfer). Should they convert to the type of the caller? In state transfer, do we transfer the data and convert it to the required type?

I assume that any marshaller is installed globally to all nodes, and not specific to the node that uses the different type, correct?

Marshalling-Upgrades.adoc
wburns: The idea is in the interceptors, the commands would be unchanged.

PR #41, opened 2026-07-03
Hot keys

tristantarrant:


HotKeys.md
```diff
@@ -0,0 +1,209 @@
...
+
+- Rectangle **area** encodes traffic volume; **color** encodes heat intensity
+  (cold blue → fire orange).
+- Click any rectangle to drill into its children (segments → keys).
+- Breadcrumb navigation at the top shows the current path in the hierarchy.
+- Bottom bar shows the hottest entity, imbalance score, total throughput,
+  and number of tracked keys.
+
+When drilling into a segment, the treemap shows individual keys with their
+ops/s and GET/PUT breakdown.
```
jabolina, line 48: At this point we don't have any tracking of ops/s

HotKeys.md
```diff
@@ -0,0 +1,209 @@
...
+
+### View 2 — Heat map grid with detail panel
+
+![Heatmap view](hotkeys-view2-heatmap.svg)
+
+**Best for:** comparing all segments (or nodes/caches/keys) side by side.
+
+- Left panel: uniform grid of cells colored by heat intensity. The selected
+  cell gets a white border highlight.
+- Right panel: detail for the selected cell — operation breakdown (GET/PUT/REMOVE
```
jabolina, line 58: We should only distinguish between reads and writes. We don't disambiguate what a write is: it could be a put, a remove, a compute, etc.

HotKeys.md
```diff
@@ -0,0 +1,209 @@
...
+  as horizontal bars), segment ownership info, and a ranked list of top keys.
+- Bottom left: a sparkline timeline (last 60 seconds) showing how the selected
```
jabolina, line 60: The tracking over time doesn't seem like something trivial to implement. We could expose the values in the metrics, and this would be possible in Grafana or something, though.

HotKeys.md
```diff
@@ -0,0 +1,209 @@
...
+
+### View 3 — Cluster topology
+
+![Topology view](hotkeys-view3-topology.svg)
+
+**Best for:** answering "which nodes are overloaded?" at a glance.
+
+- Nodes are rendered as cards with a colored border and radial glow proportional
+  to their heat. The hottest node pulses.
+- Dashed lines between nodes represent the cluster topology.
```
jabolina, line 73: I guess putting this line could cause confusion, e.g., in the SVG looks like `node-1` isn't connected to `node-3`, when in reality they are all connected.

HotKeys.md
```diff
@@ -0,0 +1,209 @@
...
+`hotKeys()` sub-builder:
+
+```java
+ConfigurationBuilder builder = new ConfigurationBuilder();
+builder.statistics()
+   .enable()
+   .hotKeys()
+      .enable()
+      .topK(100)
+      .window(60, TimeUnit.SECONDS)
```
jabolina, line 180: I need to research a bit about how to track elements in a time window. Since it is per-cache, it needs to be lightweight in terms of memory, too.

jabolina (commented):
This is aligned with what I had in mind, too. The thing I hadn't thought about was tracking within a time window. I need to get my mind around that to see how to tackle it.

I originally thought of providing the data just as a point in time. This would allow for basically all the visualizations described here, except the charts/labels of ops/s. And, if we expose the information in the metrics endpoint, I think it should be possible to create dashboards in Grafana to visualize the data over time.

jabolina:
Adding a few more aspects to bear in mind.

HeavyKeeper is inherently created to track the heavy hitters. This means that several warm keys, which draw enough load but not enough to make it into the top-k, are never tracked. We have some limitations that are:

* **Segment level**: we derive the segment from the top-k when dumping the information. This doesn't reflect the actual traffic per segment. In the warm keys, there could be segments much busier than the ones with the top-k. We could solve this with `LongAdder[numSegments]`, if we want to report something more accurate.
* **Cluster level**: we have the `HeavyKeeper` structure local to each node. In an access pattern without a defined topology, e.g., some round-robin access pattern (or embedded). A key could be warm in all nodes, never making it to the top-k in any node, but if it were a clustered structure, the key would be hot.

Part of this is because HeavyKeeper is designed for tracking heavy hitters. Users could mitigate the scenario by using a larger `k`, but that means more memory usage.

jabolina:
Another thing while talking with @wburns. We can not track node hotness based only in the top-k estimate value. We need the top-k estimation accompanied by the total hits across all keys.

Consider a 3-node cluster:

- `node-0`: 1 key with 1_000_000 reads. Total ops = 1_000_000.
- `node-1`: 10_000 keys, each with 500 reads. Total ops = 5_000_000.
- `node-2`: 8_000 keys, each with 400 reads. Total ops = 3_200_000.

Utilizing the top-k, we could measure the heatmap by _skew_ of keys, and the total hits to measure the _volume_ of access to a node. With the example above:

- **volume**: `node-1` > `node-2` > `node-0`.
- **skew**: `node-0` is `sum(top-k) / total_ops`

If we expose both the total hits AND the top-k information, a consumer would be able to generate the heatmap based on different indexes or a mixture of both.

